PR #3, opened 2024-10-09
Add specification from first labs

mhorod:


MisieqqeisiM (changes requested):
Add a section about comments

MisieqqeisiM (approved):

PR #14, opened 2024-10-11
Added arithmetic operators examples

jjaworska:


examples/incorrect/arithemtic_operators_syntax_errors.txt
mhorod, line 1: Please split this into multiple files so there's only one syntax error in each.

examples/incorrect/arithemtic_operators_with_wrong_types.txt
mhorod, line 1: Similarily with these, I think even with type errors it would be better to split them across files

mhorod (changes requested):
Also we can use `.ccp` extension now instead of `.txt`

examples/incorrect/arithemtic_operators_syntax_errors.txt
```diff
@@ -0,0 +1,13 @@
+# some syntax errors
++ 4;
```
KacperTopolski, line 2: I do not think unary plus should cause an error.

mhorod: for now we can leave it like this, let's start with binary operators only and then introduce unary operators

examples/incorrect/arithemtic_operators_syntax_errors.txt
jjaworska: Done. I also removed logical operators examples, wasn't aware it was a separate task

mhorod (approved):

PR #15, opened 2024-10-11
Added examples mixing loops and ifs

jjaworska:


examples/incorrect/mixing-loops-and-ifs-messed-up-parentheses.txt
mhorod, line 1: Please put these examples in separate files

mhorod (changes requested):
We can use `.ccp` extension

mhorod (approved):

PR #17, opened 2024-10-11
Added variable and type names examples

SebastianSpyrzewski:


mhorod (changes requested):
Please split incorrect examples into separate files and use `.ccp` file extension

mhorod (approved):

PR #18, opened 2024-10-11
added semi-cursed code example

MisieqqeisiM:


examples/correct/cursed.ccp
```diff
@@ -0,0 +1,17 @@
+let x: Unit = while true do break;
+
+let y: Int = (((((0)))));
+
+while true do while (break; true) do 1 / 0;
```
mhorod, line 5: I'd add some comment indicating what happens here, I guess the `while (break; true)` exists the outer loop?

mhorod (changes requested):

mhorod (approved):

PR #19, opened 2024-10-12
Added examples for function overloading

FilipKon13:


examples/correct/function-overloading.ccp
```diff
@@ -0,0 +1,80 @@
...
+            )
+        )
+    )
+)
+
+# shadowing/overloading in same scope
+(
+    let f = [] -> Bool => true;
+    let f = [x: Int] => Int => x;
+    let f = [] -> Int => 1;
```
mhorod, line 37: Shouldn't this be illegal? There's no way to tell which overload is called with `f[]`. Shadowing works when in nested scope we introduce function/variable with the same name, no?

examples/incorrect/function-overloading-visibility-1.ccp
```diff
@@ -0,0 +1,8 @@
+# shadowing variable does not retrieve old overloads
+let f = [] -> Bool => true;
+let f : Int = 7;
```
mhorod, line 3: My impression was that we put functions and variables int the same namespace, so that we could have first-class functions.

So the error would actually be already at this point, when we have a variable and a function with the same name?

examples/correct/function-overloading.ccp
FilipKon13: I was under the impression that shadowing works even in the same scope, ie. when declaring new variable with the same name we effectively lose access to previous one.

In this example second f is overload to first one and third f shadows first one, so after 37 we have 2 overloads.

mhorod: sounds a bit cursed, but let's go with that, we'll see what happens

mhorod (approved):

PR #23, opened 2024-10-12
Add mutual recursion examples

Rafi0221:


mhorod (approved):
I'm not really sure if that's the approach we want to take, but for me it seems most aligned with the syntax and lambdas we might want to add

PR #24, opened 2024-10-12
Added assignment examples

KacperTopolski:


examples/correct/assignment-with-function-calls.ccp
```diff
@@ -0,0 +1,11 @@
+let f = [x: Int] -> Int => x;
+
+let x: Int = 42;
+x = x + 1;
+
+let y: Int = f[x];
+y = x + f[y];
+
+x = -y;
```
mhorod, line 9: let's not have unary operators for now, they are not crucial and we can add them later to have simpler first version

PR #25, opened 2024-10-12
Added expression examples

KacperTopolski:


examples/correct/complex-expressions.ccp
```diff
@@ -0,0 +1,6 @@
+let f = [x: Int, z: Int] -> Int => x * z + 1;
+
+let x: Int = 3;
+let y: Int = 3;
+
+x = f[y = x * y + (-5) % x = f[x, x], y = -f[+-+y, -+-y]];
```
mhorod, line 6: `-f[+-+y, -+-y]`
let's not have unary operators in first version

examples/correct/operator-precedence.ccp
```diff
@@ -0,0 +1,2 @@
+let x: Int = 5 + 6 * 7 - 80 / 14;
+let y: Int = -1 + 5 -+- (7 * 7 / 8 * 9 % 5 + 10);
```
mhorod, line 2: also here - we can live without unary operators for now

mhorod (changes requested):

mhorod (approved):

Issue #28, opened 2024-10-15
Multiple variable definitions


mhorod:
One `let` can define only one variable, fix example: https://github.com/mhorod/tcs-lang-compiler/blob/0b5654633ecfb0a809bbc71ac85736bf3654a2a2/examples/incorrect/variable_definitions_multi_type_line.ccp#L3

Issue #26, opened 2024-10-15
Variable shadowing


mhorod:
Fix example: https://github.com/mhorod/tcs-lang-compiler/blob/0b5654633ecfb0a809bbc71ac85736bf3654a2a2/examples/incorrect/variable_definitions_duplicate_names.ccp#L4
We allow duplicate names - the second one shadows the previous one.

PR #30, opened 2024-10-15
Adjusted to specification

FilipKon13:


FilipKon13:
Should resolve #28 and #26

Giselus:
LGTM :thumbsup:  It's not that important to fix everything now, as we may find errors (if we still have some) when we'll try to test these examples.

Giselus (changes requested):

Giselus (approved):

PR #32, opened 2024-10-16
Add DFA equivalence checking

honvex-cool:
Added simple tools for checking whether two given states of two DFAs are equivalent
Added basic unit tests


src/test/kotlin/cacophony/automata/DFAEquivalenceTest.kt
```diff
@@ -0,0 +1,393 @@
...
+                    Pair(1, 'a') to 1,
+                    Pair(1, 'b') to 1,
+                ),
+                setOf(1),
+            )
+        assertTrue(areEquivalent(first, second))
+    }
+
+    @Test
+    fun testTwoDFAsForDifferentSimpleLanguagesButOfTheSameSizeAreNotEquivalent() {
```
mhorod, line 84: please use `snake_case` for test methods, especially when they are this long

src/test/kotlin/cacophony/automata/DFAEquivalenceTest.kt
```diff
@@ -0,0 +1,398 @@
...
+                1,
+                mapOf(
+                    Pair(1, 'a') to 1,
+                    Pair(1, 'b') to 1,
+                    Pair(2, 'a') to 1,
+                    Pair(3, 'b') to 1,
+                    Pair(4, 'a') to 2,
+                    Pair(4, 'b') to 3,
+                ),
+                setOf(),
```
Giselus, line 371: Accepting state is not _unreachable_, it just doesn't exist

src/main/kotlin/cacophony/automata/DFAHelper.kt
```diff
@@ -0,0 +1,81 @@
...
+    }
+    return object : DFAHelper<StateA, StateB> {
+        override fun areDistinguishable(
+            a: StateA,
+            b: StateB,
+        ): Boolean {
+            return distinguishable.contains(Pair(a, b))
+        }
+    }
+}
```
Giselus, line 62: I'm not a fan of the _DFAHelper_ name, it's not too descriptive and it has only one functionality so far. Maybe it should be something like _DFAEquivalenceHelper_ or something similar.
I'm also not a fan of doing everything in createHelper function, it's kind of factory/constructor, maybe we can delegate most of the code to some function like _initializeDistinguishableStates_.

Giselus (changes requested):
LGTM mostly 👍 . I don't like that all of these tests are very repetitive (create DFAs, check equivalence), but there is not too much boilerplate, so it's probably okay. Also I cannot think of any more interesting tests, so I think they're good for now.
Also test names don't have to start with 'test' I think.

src/test/kotlin/cacophony/automata/DFAEquivalenceTest.kt
honvex-cool: As discussed, I switched to using the Kotlin feature where one can enclose the test name in backticks

src/test/kotlin/cacophony/automata/DFAEquivalenceTest.kt
honvex-cool: They were supposed to be {2, 3, 4}
Fixed, test passes

src/main/kotlin/cacophony/automata/DFAHelper.kt
honvex-cool: Changed the naming and delegated the logic as suggested

Giselus (approved):

PR #33, opened 2024-10-16
Basic regex parsing

FilipKon13:
This is only basic version to allow an early feedback. 

In a final version we probably want some special characters for the set of the capital letters etc. to not write something like `(A|B|C|D|...|Z)` in the Type's regex - to be determined at the moment.


src/main/kotlin/cacophony/regex/RegexParser.kt
```diff
@@ -0,0 +1,149 @@
+package cacophony.regex
+
+import cacophony.utils.AlgebraicRegex
+import java.util.EmptyStackException
+import java.util.Stack
+
+fun parseRegex(str: String): AlgebraicRegex {
```
km1chno, line 7: im wondering if this could be a static mehod `AlgebraicRegex.fromString(str: String)`

apparently in Kotlin you do such things with [companion objects](https://kotlinlang.org/docs/object-declarations.html#companion-objects)

FilipKon13: I had same idea and found out about companion objects, however I wanted to avoid touching AlgebraicRegex class/file for the time being to not mess with concurrent work. But eventually, I agree

src/main/kotlin/cacophony/regex/RegexUtil.kt
```diff
@@ -0,0 +1,81 @@
...
+
+import cacophony.utils.AlgebraicRegex
+import cacophony.utils.AlgebraicRegex.AtomicRegex
+import cacophony.utils.AlgebraicRegex.ConcatenationRegex
+import cacophony.utils.AlgebraicRegex.StarRegex
+import cacophony.utils.AlgebraicRegex.UnionRegex
+
+class RegexSyntaxErrorException : Exception()
+
+fun algebraicRegexEquals(
```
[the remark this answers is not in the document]
FilipKon13: As above

src/main/kotlin/cacophony/regex/RegexUtil.kt
```diff
@@ -0,0 +1,81 @@
...
+        is StarRegex ->
+            if (y is StarRegex) {
+                return algebraicRegexEquals(x.internalRegex, y.internalRegex)
+            }
+        is UnionRegex ->
+            if (y is UnionRegex) {
+                return x.internalRegexes.size == y.internalRegexes.size &&
+                    x.internalRegexes.zip(y.internalRegexes).all {
+                            (a, b) ->
+                        algebraicRegexEquals(a, b)
```
SebastianSpyrzewski, line 37: Does it cover for example x=(a|b), y=(b|a)?

FilipKon13: Only for tests, it also does not work for something like (a|b|c) and (a|(b|c))

FilipKon13: Moved to test directory

src/main/kotlin/cacophony/regex/RegexParser.kt
Giselus: I think you can freely move it, as long as you don't modify existing part, it won't cause any conflicts, and even if you do, it's part of the job :x

src/main/kotlin/cacophony/regex/RegexUtil.kt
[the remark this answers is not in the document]
Giselus: Same as above

src/test/kotlin/cacophony/regex/RegexParserTest.kt
```diff
@@ -9,6 +9,58 @@ import org.junit.jupiter.api.Test
 import org.junit.jupiter.api.assertThrows
 
 class RegexParserTest {
+    private fun algebraicRegexEquals(
+        x: AlgebraicRegex,
+        y: AlgebraicRegex,
```
Giselus, line 14: Please add at least some comments that it doesn't cover some situation, like (a|b) (b|a)

src/main/kotlin/cacophony/regex/RegexParser.kt
```diff
@@ -0,0 +1,148 @@
...
+
+    fun finalize(): RegexType {
+        if (result.size != 1 || stack.isNotEmpty()) throw RegexSyntaxErrorException()
+        return result.pop()
+    }
+}
+
+private sealed class Operator {
+    abstract fun addToStack(parser: RegexParser)
+}
```
Giselus, line 68: Maybe I don't understand, but why is there Operator and StackOperator?

src/test/kotlin/cacophony/regex/RegexParserTest.kt
```diff
@@ -0,0 +1,253 @@
...
+    }
+
+    @Test
+    fun `do not recognize empty word here either`() {
+        assertThrows<RegexSyntaxErrorException> { parseRegex("abc|x()z|def") }
+    }
+
+    @Test
+    fun `mismatched parenthesis 1`() {
+        assertThrows<RegexSyntaxErrorException> { parseRegex("abc|x(z|d))ef") }
```
Giselus, line 181: You shouldn't name tests like "doing something 1/2/3...", you can add few asserts to the same test as long as it's similar case

src/main/kotlin/cacophony/regex/RegexUtil.kt
```diff
@@ -0,0 +1,49 @@
...
+import cacophony.utils.AlgebraicRegex.AtomicRegex
+import cacophony.utils.AlgebraicRegex.ConcatenationRegex
+import cacophony.utils.AlgebraicRegex.StarRegex
+import cacophony.utils.AlgebraicRegex.UnionRegex
+
+class RegexSyntaxErrorException : Exception()
+
+internal sealed class RegexType {
+    abstract fun toAlgebraicRegex(): AlgebraicRegex
+}
```
Giselus, line 13: Is it needed? Can't we use already existing Regex classes? If the interface is wrong and conversion is hard, it can be changed.

Giselus (changes requested):

src/main/kotlin/cacophony/regex/RegexParser.kt
```diff
@@ -0,0 +1,148 @@
...
+                    parser.pushSymbol(c)
+                }
+            }
+        }
+    }
+    return parser.finalize().toAlgebraicRegex()
+}
+
+private class RegexParser {
+    val result = Stack<RegexType>()
```
Giselus, line 41: Stack is outdated, use ArrayDeque instead.

src/main/kotlin/cacophony/regex/RegexParser.kt
```diff
@@ -0,0 +1,148 @@
...
+
+fun parseRegex(str: String): AlgebraicRegex {
+    val parser = RegexParser()
+    val regex = "($str)"
+    var specialCharacter = false
+    var lastEscaped = -1
+    regex.indices.forEach {
+        val c = regex[it]
+        if (specialCharacter) {
+            if (c !in SPECIAL_CHARACTER_MAP.keys) throw RegexSyntaxErrorException()
```
Giselus, line 15: Maybe we can add more descriptive errors? Like `throw RegexSyntaxErrorException("Expected special character, but got $c at position x")`. Similarly in other places.

Giselus (changes requested):

src/main/kotlin/cacophony/regex/RegexParser.kt
AHGIJMKLKKZNPJKQR: You can declare an [extension function](https://kotlinlang.org/docs/extensions.html#extension-functions) to get this syntax without modifying `AlgebraicRegex`

src/main/kotlin/cacophony/regex/RegexUtil.kt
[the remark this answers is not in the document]
AHGIJMKLKKZNPJKQR: The Same As Above

src/main/kotlin/cacophony/regex/RegexParser.kt
FilipKon13: Operator is anything that acts as operator in regex syntax (implicite concat, |, (, ) and *. However, ) and * never actually land on stack of operators in parser (bacause they are resolved immediately), therefore do not need priority. StackOperators can land on the stack, therefore need they priority for additional logic

src/main/kotlin/cacophony/regex/RegexUtil.kt
FilipKon13: Union and Concat in AlgebraicRegex declare its internals as valargs, therefore does not allow for cheap addition to them (which is used as simple optimalization to reduce nesting of resulting AlgebraicRegex) - as they should btw, AlgebraicRegex should be somehow immutable object in my opiniom.

Other thing is, RegexType is thought as internal to the parser - in theory there may be new operators (eg. ? or []) which AlgebraicRegex does not need, or additional info (eg. position in starting stringi for error printing, as below) needed only for parser.

Think of RegexType as Builder type for AlgebraicRegex

src/main/kotlin/cacophony/regex/RegexParser.kt
FilipKon13: Added `AlgebraicRegex.fromString(str: String)` function

src/main/kotlin/cacophony/regex/RegexUtil.kt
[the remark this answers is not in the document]
FilipKon13: Moved both functions to corresponding test file

src/test/kotlin/cacophony/regex/RegexParserTest.kt
FilipKon13: Done

src/test/kotlin/cacophony/regex/RegexParserTest.kt
FilipKon13: Renamed

src/main/kotlin/cacophony/regex/RegexParser.kt
FilipKon13: Done

src/main/kotlin/cacophony/regex/RegexParser.kt
Giselus: So I think it would be nice to just have Operator with priority, this priority will be just anything for ) and *. Code would be cleaner I think

Giselus (approved):

src/main/kotlin/cacophony/regex/RegexParser.kt
FilipKon13: Having ) as something which can be on stack would require handling it (probably by throwing) in every type matching on top stack element, this way we have static guarantees. Removed this class anyway, Star can be now on stack, right parenthesis is handled separately (I guess this way is also correct by static inspection? Dunno if cleaner)

PR #34, opened 2024-10-18
Add initial nfa to dfa implementation

AHGIJMKLKKZNPJKQR:


src/main/kotlin/cacophony/automata/Determinization.kt
```diff
@@ -0,0 +1,87 @@
...
+    symbol: Char,
+): Set<StateType> =
+    this.epsilonClosure(states).also { closure ->
+        closure.addAll(this.epsilonClosure(closure.flatMap { this.getProductions(it, symbol) }))
+    }
+
+private class SimpleDFA<StateType>(
+    private val start: StateType,
+    private val prod: Map<Pair<StateType, Char>, StateType>,
+    private val accept: Set<StateType>,
```
Giselus, line 71: Maybe we can create some IntDFA or SimpleDFA<StateType>, and move it somewhere else, perhaps someone would like to use it as well (during minimalization).

src/main/kotlin/cacophony/automata/Determinization.kt
```diff
@@ -0,0 +1,87 @@
...
+    val acceptingStates = dfaProductions.keys.filter { it.any { nfa.isAccepting(it) } }.map { setToInt[it]!! }.toSet()
+    val productions =
+        dfaProductions
+            .flatMap { (state, edges) -> edges.map { Pair(setToInt[state]!!, it.key) to setToInt[it.value]!! } }
+            .toMap()
+    return SimpleDFA(setToInt[startingState]!!, productions, acceptingStates)
+}
+
+private fun <StateType> NFA<StateType>.epsilonClosure(states: Iterable<StateType>): MutableSet<StateType> {
+    var queue = states.toMutableList()
```
Giselus, line 42: Isn't there some Queue/Deque?

src/main/kotlin/cacophony/automata/Determinization.kt
```diff
@@ -0,0 +1,87 @@
...
+
+private fun <StateType> determinize(
+    nfa: NFA<StateType>,
+    charset: Iterable<Char>,
+): DFA<Int> {
+    val startingState = setOf(nfa.getStartingState())
+    var list = mutableListOf(startingState)
+    var dfaProductions = mutableMapOf<Set<StateType>, MutableMap<Char, Set<StateType>>>()
+
+    while (!list.isEmpty()) {
```
Giselus, line 24: Are we missing pushing to list? And the list could also be some queue object.

Giselus (changes requested):

src/main/kotlin/cacophony/automata/Determinization.kt
AHGIJMKLKKZNPJKQR: Very interesting BFS implementation indeed

src/main/kotlin/cacophony/automata/Determinization.kt
AHGIJMKLKKZNPJKQR: Switched to `ArrayDeque`, I kind of assumed `MutableList` would be a linked list

src/main/kotlin/cacophony/automata/Determinization.kt
AHGIJMKLKKZNPJKQR: I moved this DFA to `DFA.kt`, though I'm not sure if anyone will make any use of it since they're working simultaneously

src/test/kotlin/cacophony/automata/DeterminizationTest.kt
```diff
@@ -0,0 +1,187 @@
...
+
+            override fun getAcceptingState() = accepting
+
+            override fun isAccepting(state: Int) = state == accepting
+
+            override fun getAllStates() = allStates
+        }
+
+    @Test
+    fun `Empty language NFA maps to empty language DFA`() {
```
Giselus, line 45: I think we started with naming test functions with lowercase first letter, so maybe it's the way to go, to keep one style

src/main/kotlin/cacophony/automata/DFA.kt
```diff
@@ -20,3 +20,30 @@ interface DFA<DFAState> {
     // Returned value is map accepting current state and symbol, and returning new state, which may not exist.
     fun getProductions(): Map<Pair<DFAState, Char>, DFAState>
 }
+
+public data class SimpleDFA<StateType>(
```
Giselus, line 24: If you created it maybe you add at least one test to see if at least getAllStates is correct, it's like the only place with some logic.

Giselus (approved):
LGTM 👍 Approving so you don't have to wait for me later after resolving last two comments.

PR #36, opened 2024-10-18
Added token categories and regex description

SebastianSpyrzewski:


src/main/kotlin/cacophony/regex/RegexStrings.kt
```diff
@@ -0,0 +1,88 @@
...
+            TokenCategorySpecific.OPERATOR_LOGICAL_NOT to """!""",
+            // and the others
+            TokenCategorySpecific.LITERAL to """-\d(\d)*|\d(\d)*|true|false""",
+            TokenCategorySpecific.TYPE_IDENTIFIER to """\u\w*""",
+            TokenCategorySpecific.VARIABLE_IDENTIFIER to """(\l|_)\w*""",
+            TokenCategorySpecific.WHITESPACE to """(\n|\r|\t| )*""",
+            TokenCategorySpecific.COMMENT to """#\N*""",
+        )
+
+    private val specialCharacterMap =
```
Giselus, line 69: Doesn't it clash with Filip's implementation? I see that he already has special symbols for \l, \u. I truly don't know what's the best place to have this stuff, maybe in Filip's part, as you don't have any logic around it at all? I'm not sure tho. And make sure that you agree on this special characters, as I don't see \d and \w in Filip's code.

Giselus (approved):
And if I understand, we want to delete general or specific categories later? If so, please add some comment about this. I approve in advance, so you don't have to wait later

src/test/kotlin/cacophony/regex/RegexStringsTest.kt
```diff
@@ -0,0 +1,26 @@
+package cacophony.regex
+
+import cacophony.token.TokenCategoryGeneral
+import cacophony.token.TokenCategorySpecific
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+internal class RegexStringsTest {
+    @Test
```
Giselus, line 9: And I think we agreed to name test functions with names like: `test special character regex`, maybe without word 'test' (you can check in other branches)

Giselus (approved):

src/main/kotlin/cacophony/regex/RegexStrings.kt
FilipKon13: My oversight, \w and \d will be added.

PR #40, opened 2024-10-19
Add StringInput and FileInput

Rafi0221:


src/main/kotlin/cacophony/utils/FileInput.kt
```diff
@@ -0,0 +1,91 @@
...
+        // Used to encode Location.
+        var linesLengthSum = 0
+        lineBeginPos.add(0)
+        for (line in lines) {
+            linesLengthSum += line.length
+            lineBeginPos.add(linesLengthSum)
+        }
+    }
+
+    private fun setNextPosition() {
```
Giselus, line 28: Maybe advancePosition or something similar?

src/main/kotlin/cacophony/utils/FileInput.kt
```diff
@@ -0,0 +1,91 @@
...
+
+    // Location is encoded as the position in the string formed by the concatenation of all lines.
+    private fun encodeLocation(
+        lineInd: Int,
+        pos: Int,
+    ): Location {
+        return Location(lineBeginPos[lineInd] + pos)
+    }
+
+    private fun decodeLocation(loc: Location): Pair<Int, Int> {
```
Giselus, line 54: What is someone passes negative or astronomically big index?

src/main/kotlin/cacophony/utils/FileInput.kt
```diff
@@ -0,0 +1,91 @@
...
+        return encodeLocation(curLineInd, curPos)
+    }
+
+    override fun setLocation(loc: Location) {
+        val (lineInd, pos) = decodeLocation(loc)
+        curLineInd = lineInd
+        curPos = pos
+    }
+
+    override fun locationToString(loc: Location): String {
```
Giselus, line 86: I would also add some function to print content between two locations or from one location to the end of the line? Would be helpful to show errors.

src/main/kotlin/cacophony/utils/StringInput.kt
```diff
@@ -0,0 +1,34 @@
+package cacophony.utils
+
+class StringInput(private val text: String) : Input {
+    // current position of the cursor in string
```
Giselus, line 4: Let's be consistent and start comments with uppercase and place dot at the end.

src/test/kotlin/cacophony/utils/FileInputTest.kt
```diff
@@ -0,0 +1,311 @@
+package cacophony.utils
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+import java.io.File
+
+class FileInputTest {
+    private fun createFileWithContent(content: String): String {
+        val file = File.createTempFile("test_file", ".cac")
```
Giselus, line 9: I think you don't have to specify filenames, as it will get some random number appended. But it's okay.

src/test/kotlin/cacophony/utils/FileInputTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        file.deleteOnExit()
+        file.writeText(content)
+        return file.path
+    }
+
+    @Test
+    fun next() {
```
Giselus, line 16: I don't like the naming of these test functions (next(), peek(), etc.), short name seems like test helper function, maybe "next() simple" or something like that would be better.

src/test/kotlin/cacophony/utils/FileInputTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        val path = createFileWithContent("abcde")
+        val input = FileInput(path)
+
+        assertEquals('a', input.next())
+        assertEquals('b', input.next())
+        assertEquals('c', input.next())
+        assertEquals('d', input.next())
+        assertEquals('e', input.next())
+        // not checking first character after last - it is '\n'
```
Giselus, line 25: I don't understand what you meant here, can you elaborate? Why we don't check the '\n' character?

Giselus (changes requested):
Looks good in general, but I feel like tests are a little too complicated? I didn't try to read them too carefully, but they doesn't seem like minimal examples, maybe we can test the same functionality with less lines? It's just a suggestion, not requirement.

src/test/kotlin/cacophony/utils/FileInputTest.kt
Rafi0221: I manually add '\n' to the end of each line, even if it is not present in the file. This is kind of an implementation detail, so there is no assert for that.

src/main/kotlin/cacophony/utils/FileInput.kt
Rafi0221: Fixed, but I think you shouldn't create Location on your own, outside of Input

Rafi0221:
> Looks good in general, but I feel like tests are a little too complicated? I didn't try to read them too carefully, but they doesn't seem like minimal examples, maybe we can test the same functionality with less lines? It's just a suggestion, not requirement.

Yes, some tests are complicated, but I felt like functionalities can greatly interact with each other. The class itself should be easy to debug, regardless.

src/main/kotlin/cacophony/utils/FileInput.kt
Giselus: You shouldn't, but what if we add modules and we'll parse two files and somehow mix current locations? Or something other. It doesn't matter that we shouldn't do something, we still should guard ourselves from it.

Giselus (approved):

PR #41, opened 2024-10-19
Added regex to NFA conversion

jjaworska:


src/main/kotlin/cacophony/automata/NFAConstructor.kt
```diff
@@ -0,0 +1,135 @@
+package cacophony.automata
+
+import kotlin.collections.emptyMap
+import cacophony.utils.AlgebraicRegex
+
+fun buildNFAFromRegex(regex: AlgebraicRegex): NFA<Int> = buildSimpleNFAFromRegex(regex)
+
+// implementation relies on the fact that the initial state will always have number 0
```
Giselus, line 8: Please start comment with uppercase letter and end it with dot to stay consistent with other PRs.

src/test/kotlin/cacophony/automata/NFAConstructorTest.kt
```diff
@@ -0,0 +1,114 @@
...
+            var states = epsClosure(setOf(nfa.getStartingState()))
+            for (symbol in str) {
+                states = step(states, symbol)
+            }
+            return (nfa.getAcceptingState() in states)
+        }
+    }
+
+    @Test
+    fun `Simple atomic regex test`() {
```
Giselus, line 41: We also start test function names from lowercase letter.

src/main/kotlin/cacophony/automata/NFAConstructor.kt
Giselus, line 1: Looks good, but I'm not sure about this file name. There is already simpleDFA class, and it's in the DFA.kt. It doesn't have to be like that, but usally filename is the same as main interface/class name, and we don't have anything like NFAConstructor. Maybe we can move buildNFAFromRegex to NFA.kt and rename this file to simpleNFA.kt? Just suggestion, probably can be done differently.

Giselus (changes requested):
Looks mainly good, only cosmetic changes. But please, fix the build, as it's currently failing due to linter.

src/main/kotlin/cacophony/automata/NFAConstructor.kt
```diff
@@ -0,0 +1,135 @@
...
+                epsProds.getOrPut(sub.getAcceptingState()) { mutableListOf() }.add(finalState)
+            }
+            SimpleNFA(
+                initState,
+                prods,
+                epsProds,
+                finalState)
+        }
+        is AlgebraicRegex.ConcatenationRegex -> {
+            var prods: Map<Pair<Int, Char>, List<Int>> = mapOf()
```
Giselus, line 43: Maybe it should be mutableMap?

Giselus (changes requested):

src/main/kotlin/cacophony/automata/NFAConstructor.kt
jjaworska: Following the convention which was chosen for DFA's, I moved the SimpleNFA class to the NFA.kt, but left the implementation in a separate file, didn't really have a name better idea

Giselus (approved):

PR #43, opened 2024-10-19
Make basic test framework for examples

### LINKED ISSUE #42: Create test framework for examples

mhorod:
Resolves #42


mhorod:
There are few depenedncies on other here like FileInput and Lexer itself, so we should imo we can merge this one at the end when we have the whole process configured

src/test/kotlin/cacophony/examples/ExampleRunner.kt
```diff
@@ -0,0 +1,78 @@
+package cacophony.examples
+
+import cacophony.utils.Input
+import cacophony.utils.Location
+import java.io.File
+import java.nio.file.Files
+import java.nio.file.Path
+import kotlin.io.path.isRegularFile
+
+interface Runner {
```
Giselus, line 10: Maybe still ExampleRunner?

src/test/kotlin/cacophony/examples/ExampleRunner.kt
```diff
@@ -0,0 +1,78 @@
...
+        input: Input,
+        diagnostics: TestDiagnostics,
+    )
+}
+
+interface DiagnosticsAssertion {
+    fun check(diagnostics: TestDiagnostics)
+}
+
+class FileInput(path: Path) : Input {
```
Giselus, line 21: FileInput is already merged, please change.

examples/incorrect/assignment/descriptions.json
```diff
@@ -0,0 +1,34 @@
+{
+  "examples": [
+    {
+      "path": "assignment-incorrect-types.cac",
+      "errors": {
+        "lexerErrors": false
+      }
+    },
+    {
+      "path": "descriptions.json",
```
Giselus, line 10: I guess this file shouldn't contain it's own description?

Giselus (changes requested):
I'm not a fan of current solution. We should have more precise error, not only information if it occurs or not.

Giselus:
I guess there will be few more things to fix, but we'll get back to it after other things are merged.

examples/incorrect/assignment/descriptions.json
mhorod: Nice catch, I didn't even notice, will fix

mhorod:
> I'm not a fan of current solution. We should have more precise error, not only information if it occurs or not.

Yeah, I agree, although we didn't really discuss error handling. We should  get back to this when lexer gets implemented

Giselus (approved):

PR #44, opened 2024-10-19
DFA minimalization

MisieqqeisiM:


src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
```diff
@@ -0,0 +1,101 @@
...
+    fun getPreimages(states: Set<DFAState>): Map<Char, Set<DFAState>> =
+        states
+            .mapNotNull { state -> cache[state]?.let { state to it } }
+            .flatMap { it.second.entries }
+            .flatMap { entry -> entry.value.map { entry.key to it } }
+            .groupBy({ it.first }, { it.second })
+            .mapValues { it.value.toSet() }
+}
+
+fun <DFAState> getDFAForEmptyLanguage(): DFA<DFAState> = TODO("DFA interface does not allow empty languages")
```
Giselus, line 26: How it doesn't allow it? Isn't it enough to create DFA without accepting states?

src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
```diff
@@ -0,0 +1,101 @@
...
+    val fullDFA = this
+
+    return createDFA(
+        fullDFA.getStartingState(),
+        retain.filter { fullDFA.isAccepting(it) }.toSet(),
+        productions,
+    )
+}
+
+fun <DFAState> createDFA(
```
Giselus, line 73: Cannot we use SimpleDFA?

src/test/kotlin/cacophony/automata/minimalization/DFAMinimalizationTest.kt
```diff
@@ -0,0 +1,177 @@
...
+            states.random(random),
+            states.filter { random.nextDouble() < 0.2 }.toSet(),
+            (0..<ceil(density * n * symbols.length).toInt()).associate {
+                states.random(random) via symbols.random(random) to states.random(random)
+            },
+        )
+    }
+
+    @Test
+    fun `Random DFAs`() {
```
Giselus, line 174: I'm not sure about this, imo tests shouldn't be non-deterministic. If implementation is wrong it can randomly fail on build not related to this. We can leave it now and discuss it on Tuesday.

src/test/kotlin/cacophony/automata/minimalization/PartitionRefinementTest.kt
```diff
@@ -0,0 +1,186 @@
...
+    @Test
+    fun `getAllPartitions works after two refinements`() {
+        val pr = initRefinement(5)
+        pr.refine(listOf(0, 2, 3, 4))
+        pr.refine(listOf(3))
+        assertEquals(part[0, 2, 4][1][3], pr.getAllPartitions().toSet())
+    }
+
+    @Test
+    fun `getAllPartitions works after three refinements`() {
```
Giselus, line 38: You can do things like that in one test function. No need to create things like that

src/test/kotlin/cacophony/automata/minimalization/DFAMinimalizationTest.kt
KacperTopolski: Tests are deterministic as they use fixed seeds. 

java.util.Random guarantees this https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/Random.html

Kotlin's random uses slightly weaker wording: "Note: Future versions of Kotlin may change the algorithm of this seeded number generator so that it will return a sequence of values different from the current one for a given seed." I personally do not think this is a problem, but I can change the random to java one if needed.

src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
KacperTopolski: Maybe I misunderstood the spec, but I thought that non-null dead states are not allowed? This is why I asked to make DFA.getStartingState() return DFAState? instead of DFAState.

src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
KacperTopolski: SimpleDFA was not merged till yesterday and we did not know of it's existence. Fixed.

Giselus (approved):

PR #57, opened 2024-10-22
Add new interfaces for grammars and parser

### LINKED ISSUE #56: Add interfaces for Grammar and Parser

km1chno:
Resolves #56 

Adds new interfaces and function declarations to be implemented in the following week

### DIFF
--- a/src/main/kotlin/cacophony/grammars/ParseTree.kt
+++ b/src/main/kotlin/cacophony/grammars/ParseTree.kt
@@ -3,14 +3,16 @@ package cacophony.grammars
 import cacophony.token.Token
 import cacophony.utils.Location
 
-sealed class ParseTree<Symbol : Enum<Symbol>> {
-    class Leaf<Symbol : Enum<Symbol>>(
-        val token: Token<Symbol>,
-    ) : ParseTree<Symbol>()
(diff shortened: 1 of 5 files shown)


src/main/kotlin/cacophony/grammars/ParseTree.kt
```diff
@@ -0,0 +1,16 @@
...
+import cacophony.token.Token
+import cacophony.utils.Location
+
+sealed class ParseTree<Symbol : Enum<Symbol>> {
+    class Leaf<Symbol : Enum<Symbol>>(
+        val token: Token<Symbol>,
+    ) : ParseTree<Symbol>()
+
+    class Branch<Symbol : Enum<Symbol>>(
+        val range: Pair<Location, Location>,
```
Giselus, line 12: I think it may be part of any ParseTree (both Branch and Leaf). Leaf already has this info, but it would be easier to parse subtrees, as you don't have to branch on ParseTree type.

Giselus (approved):
LGTM 👍

src/main/kotlin/cacophony/grammars/ParseTree.kt
km1chno: ok I agree, this is not large cost to duplicate this

Issue #58, opened 2024-10-22
Refactor I


km1chno:
List of things to do:

- reconsider linter config, the are some things that are irritating to some people
- think of a convention for naming generic parameters and rename them, currently we don't have one. In several places they have suffix `Type`, but e.g. [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html#s5.2.8-type-variable-names) proposes suffix `T`

Giselus:
Change test productions like this: `Pair(1, 'a') to 2` -> `1 via 'a' to 2`

FilipKon13:
Move utility functions (`via`, `buildDFAFromRegex` etc.) from `DFAUtils.kt` outside of minimalization directory

AHGIJMKLKKZNPJKQR:
Improve diagnostics to be more flexible (https://github.com/mhorod/cacophony/pull/70#discussion_r1818735313)

MisieqqeisiM:
Create a superclass for all custom exceptions thrown during compilation

FilipKon13:
Resolve #27

PR #59, opened 2024-10-23
DFA/AlgebraicRegex new generic params

Giselus:


src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
```diff
@@ -72,7 +77,10 @@ fun <DFAState> DFA<DFAState>.withStates(retain: Set<DFAState>): DFA<DFAState> {
     return SimpleDFA(
         fullDFA.getStartingState(),
         productions,
-        retain.filter { fullDFA.isAccepting(it) }.toSet(),
+        retain.mapNotNull {
+            val result = fullDFA.result(it)
+            if (result != null) it to result else null
+        }.toMap(),
     )
 }
```
FilipKon13, line 86: I think we should also generalize infix `via` operator to also work with any `Atom`

FilipKon13: And maybe move this operator outside of minimalization directory, but this can be done during refactoring

src/main/kotlin/cacophony/automata/DFA.kt
```diff
@@ -1,49 +1,54 @@
 package cacophony.automata
 
-interface DFA<DFAState> {
+interface DFA<DFAState, AtomType, ResultType> {
```
km1chno, line 3: in the current state, all these 3 parameters should be non nullable, right? So we could declare

```kotlin
interface DFA<DFAState : Any, AtomType : Any, ResultType : Any> {
```

Although it looks like a frustrating task to add it everywhere

You can do it or add it to list of things to do in the refactor https://github.com/mhorod/cacophony/issues/58.
We could also replace these params with interfaces, it's getting a bit too much of them in my opinion (also as part of the refactor)

src/main/kotlin/cacophony/automata/Determinization.kt
```diff
@@ -4,29 +4,29 @@ import kotlin.collections.mutableMapOf
...
         },
     )
 
-private fun <StateType> determinize(
-    nfa: NFA<StateType>,
-    charset: Iterable<Char>,
-): DFA<Int> {
+private fun <StateType, AtomType> determinize(
+    nfa: NFA<StateType, AtomType>,
+    atomset: Iterable<AtomType>,
```
km1chno, line 18: `atoms`?

src/main/kotlin/cacophony/automata/Determinization.kt
```diff
@@ -40,7 +40,7 @@ private fun <StateType> determinize(
         createdStates
             .filter { it.any { nfa.isAccepting(it) } }
             .map { setToInt[it]!! }
-            .toSet()
+            .associate { it to true }
```
km1chno, line 43:
```kotlin
val acceptingStates =
    createdStates
        .filter { it.any { state -> nfa.isAccepting(state) } }
        .map { setToInt[it]!! }
        .associateWith { true }
```

src/main/kotlin/cacophony/grammars/Analysis.kt
```diff
@@ -2,37 +2,34 @@ package cacophony.grammars
...
+): Map<DFAStateReference<StateType, *, SymbolType>, Collection<SymbolType>> {
     TODO("Not implemented")
 }
 
 data class AnalyzedGrammar<StateType, SymbolType>(
     val syncSymbolTypes: Collection<SymbolType>,
-    val automata: Map<SymbolType, DFA<StateType>>,
-    // TODO: Uncomment after #46
-    // val automata: Map<SymbolType, DFA<StateType, Atom, SymbolType>>
+    val automata: Map<SymbolType, DFA<StateType, *, SymbolType>>,
```
km1chno, line 30: this is not the valid type, it should be

```kotlin
val automata: Map<SymbolType, DFA<StateType, SymbolType, Production<SymbolType>>>
```

because these the alphabet of these automata are symbols in our grammars (`SymbolType`), and the result is from which production the match was found (so the result is just `Production<SymbolType>`)

km1chno (changes requested):
Looks good overall, some nitpicks and one thing with type of automata in `AnalyzedGrammar`

It is a bit stupid that in Boolean DFA states with result equal to `false` are accepting 😄 But it is what it is

src/main/kotlin/cacophony/automata/DFAEquivalenceHelper.kt
```diff
@@ -36,21 +36,21 @@ fun <StateA, StateB> createDFAEquivalenceHelper(
...
     val symbols = getSymbols(dfaA.getProductions()) union getSymbols(dfaB.getProductions())
     val invA = invertProductions(dfaA, symbols)
     val invB = invertProductions(dfaB, symbols)
     val distinguishable = mutableSetOf<Pair<StateA?, StateB?>>()
     for (a in dfaA.getAllStates())
         for (b in dfaB.getAllStates())
-            if (dfaA.isAccepting(a) != dfaB.isAccepting(b)) {
+            if (dfaA.result(a) != dfaB.result(b)) {
                 distinguishable.add(Pair(a, b))
             }
```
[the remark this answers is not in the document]
Giselus: It's unreadable, my conscience doesn't allow me to change it like that.

src/main/kotlin/cacophony/automata/DFA.kt
Giselus: Added comment to refactor task

src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
Giselus: Changed to any Atom and added comment to refactor

km1chno (approved):
good job!

PR #60, opened 2024-10-23
Joining DFAs

FilipKon13:
Initial implementation to allow early feedback. Changes in `DFA.kt` were done only to allow project to compile and will get reverted after merging #59

### DIFF
--- a/src/main/kotlin/cacophony/automata/DFAJoin.kt
+++ b/src/main/kotlin/cacophony/automata/DFAJoin.kt
@@ -2,17 +2,17 @@ package cacophony.automata
 
 import cacophony.automata.minimalization.via
 
-// Joins given list of automata into single DFA with accepting states marked
-// with Result corresponding to source DFAs.
-// Throws if any state would be marked by many results, i.e. when any two DFAs accept same word.
+// Joins given list of the automata into a single DFA with accepting states marked
+// with a Result corresponding to the source DFAs.
+// Throws if any state would be marked by many results, i.e. when any two DFAs accept the same word.
 fun <State, Atom, Result> joinAutomata(
     automataAndResults: List<Pair<GenericDFA<State, Atom, *>, Result>>,
 ): GenericDFA<Int, Atom, Result?> {
     if (automataAndResults.isEmpty()) throw IllegalArgumentException("Provided empty list of automaton")
     val automata = automataAndResults.unzip().first
 
-    // Get result for given state or null if it is not accepting.
-    // Throws if result is ambiguous.
+    // Get result for the given state or null if it is not accepting.
+    // Throws if the result is ambiguous.
     val resultForState = fun(states: List<State?>): Result? {
         val accept = (automataAndResults zip states).filter { (pair, state) -> state != null && pair.first.isAccepting(state) }
         if (accept.size > 1) throw IllegalArgumentException("Provided automata are ambiguous")
@@ -20,17 +20,17 @@ fun <State, Atom, Result> joinAutomata(
         return res.first.second
     }
 
-    // Apply transition to every element of list.
+    // Apply the transition to every element of list.
     val transition = fun(
         states: List<State?>,
         atom: Atom,
     ): List<State?> = (automata zip states).map { (automaton, state) -> state?.let { automaton.getProduction(it, atom) } }
 
-    // Parts of final automaton.
+    // Parts of the final automaton.
     val productions: MutableMap<Pair<Int, Atom>, Int> = mutableMapOf()
     val results: MutableMap<Int, Result?> = mutableMapOf()
 
-    // Preparation to graph search.
+    // Preparation to the graph search.
     val viableTransitions = automata.flatMap { it.getProductions().keys.map { it.second } }.toSet()
     val oldStateToNew: MutableMap<List<State?>, Int> = mutableMapOf()
     val queue = ArrayDeque<Pair<List<State?>, Int>>()
@@ -40,7 +40,7 @@ fun <State, Atom, Result> joinAutomata(
     oldStateToNew[startingState] = stateCounter++
     while (!queue.isEmpty()) {
         val (currentState, currentName) = queue.removeFirst()
-        // Throws if currentState has ambiguous Result
+        // Throws if currentState has an ambiguous Result
         results[currentName] = resultForState(currentState)
         for (atom in viableTransitions) {
             val nextState = transition(currentState, atom)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/cacophony/automata/DFAJoin.kt
```diff
@@ -0,0 +1,83 @@
+package cacophony.automata
+
+import cacophony.automata.minimalization.via
+
+// Joins given list of the automata into a single DFA with accepting states marked
+// with a Result corresponding to the source DFAs.
+// Throws if any state would be marked by many results, i.e. when any two DFAs accept the same word.
+fun <State, Atom, Result> joinAutomata(
```
km1chno, line 8: in #59 we kind of embraced the convention of adding suffix `Type` to type arguments, so I'd stick to that, so
`StateType`, `AtomType` and `ResultType`

src/main/kotlin/cacophony/automata/DFAJoin.kt
```diff
@@ -0,0 +1,83 @@
...
+    if (automataAndResults.isEmpty()) throw IllegalArgumentException("Provided empty list of automaton")
+    val automata = automataAndResults.unzip().first
+
+    // Get result for the given state or null if it is not accepting.
+    // Throws if the result is ambiguous.
+    val resultForState = fun(states: List<State?>): Result? {
+        val accept = (automataAndResults zip states).filter { (pair, state) -> state != null && pair.first.isAccepting(state) }
+        if (accept.size > 1) throw IllegalArgumentException("Provided automata are ambiguous")
+        val res = accept.firstOrNull() ?: return null
+        return res.first.second
```
km1chno, line 20:
```kotlin
return accept.firstOrNull()?.first?.second
```

src/main/kotlin/cacophony/automata/DFAJoin.kt
```diff
@@ -0,0 +1,83 @@
+package cacophony.automata
+
+import cacophony.automata.minimalization.via
+
+// Joins given list of the automata into a single DFA with accepting states marked
+// with a Result corresponding to the source DFAs.
+// Throws if any state would be marked by many results, i.e. when any two DFAs accept the same word.
+fun <State, Atom, Result> joinAutomata(
+    automataAndResults: List<Pair<GenericDFA<State, Atom, *>, Result>>,
+): GenericDFA<Int, Atom, Result?> {
```
km1chno, line 10: with @Giselus we decided that `Result` param in DFA will be non-nullable, and instead we recognize that state is not accepting iff it does not exist as key of the `results` map

src/main/kotlin/cacophony/automata/DFAJoin.kt
```diff
@@ -0,0 +1,83 @@
...
+        states: List<State?>,
+        atom: Atom,
+    ): List<State?> = (automata zip states).map { (automaton, state) -> state?.let { automaton.getProduction(it, atom) } }
+
+    // Parts of the final automaton.
+    val productions: MutableMap<Pair<Int, Atom>, Int> = mutableMapOf()
+    val results: MutableMap<Int, Result?> = mutableMapOf()
+
+    // Preparation to the graph search.
+    val viableTransitions = automata.flatMap { it.getProductions().keys.map { it.second } }.toSet()
```
km1chno, line 34: please rename the second `it` to avoid shadowing

src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
```diff
@@ -76,4 +79,9 @@ fun <DFAState> DFA<DFAState>.withStates(retain: Set<DFAState>): DFA<DFAState> {
     )
 }
 
-infix fun <DFAState> DFAState.via(label: Char): Pair<DFAState, Char> = Pair(this, label)
+infix fun <DFAState, Atom> DFAState.via(label: Atom): Pair<DFAState, Atom> = Pair(this, label)
+
+// Some utility functions, to not write whole pipeline each time
+fun buildDFAFromRegex(regex: AlgebraicRegex) = determinize(buildNFAFromRegex(regex)).minimalize()
+
+fun buildDFAFromRegex(regex: String) = buildDFAFromRegex(AlgebraicRegex.fromString(regex))
```
km1chno, line 87: I feel like this should not be in `minimalization` package 😄

src/main/kotlin/cacophony/automata/DFAJoin.kt
```diff
@@ -0,0 +1,83 @@
...
+                oldStateToNew.getOrPut(nextState) {
+                    queue.addLast(nextState to stateCounter)
+                    stateCounter++
+                }
+            productions[currentName via atom] = nextName
+        }
+    }
+
+    // TODO: Do we need to clear dead states?
+    // TODO: minimalize?
```
km1chno, line 57: yes! after #59 you should be able to minimalize

src/main/kotlin/cacophony/automata/DFAJoin.kt
```diff
@@ -0,0 +1,83 @@
...
+            val nextName =
+                oldStateToNew.getOrPut(nextState) {
+                    queue.addLast(nextState to stateCounter)
+                    stateCounter++
+                }
+            productions[currentName via atom] = nextName
+        }
+    }
+
+    // TODO: Do we need to clear dead states?
```
km1chno, line 56: minimalization should do it

km1chno (changes requested):
nice! left some initial thoughts

src/main/kotlin/cacophony/automata/DFAJoin.kt
FilipKon13: Coverage cries that second null check never goes to false : ' )

src/main/kotlin/cacophony/automata/DFAJoin.kt
FilipKon13: That was my first version also, changed when I saw @Giselus PR. Will change it back

src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
FilipKon13: Yeah, placed it there so that it can be moved along with 'via' during refctor

src/main/kotlin/cacophony/automata/DFAJoin.kt
km1chno: alright then it can stay like this :D

src/main/kotlin/cacophony/automata/minimalization/DFAUtils.kt
km1chno: if you want to leave something for the refactor, please write a comment in https://github.com/mhorod/cacophony/issues/58

km1chno (approved):
looks good to me <3, please just write a comment in https://github.com/mhorod/cacophony/issues/58 about extracting these utils during refactor

PR #62, opened 2024-10-26
Add AST class hierarchy and priority description

SebastianSpyrzewski:
I don't know if it should be implemented like this, so it can be easily changed in the future

### DIFF
--- a/src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
+++ b/src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
@@ -60,9 +60,19 @@ sealed class Expression(val range: Pair<Location, Location>) { // everything in 
         vararg val arguments: Expression,
     ) : Expression(range)
 
-    class Literal(
+    sealed class Literal(
         range: Pair<Location, Location>,
-    ) : Expression(range)
+    ) : Expression(range) {
+        class IntLiteral(
+            range: Pair<Location, Location>,
+            val value: Int,
+        ) : Literal(range)
+
+        class BoolLiteral(
+            range: Pair<Location, Location>,
+            val value: Boolean,
+        ) : Literal(range)
+    }
 
     // expression in parenthesis
     class ParenthesisGroup(
--- a/src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
+++ b/src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
@@ -7,7 +7,8 @@ sealed class Expression(val range: Pair<Location, Location>) { // everything in 
         range: Pair<Location, Location>,
     ) : Expression(range)
 
-    class Type( // should Type also inherit expression? if so, can we do for example "; Bool;" in our language?
+    // should Type also inherit expression? if so, can we do for example "; Bool;" in our language?
+    class Type(
         range: Pair<Location, Location>,
     ) : Expression(range)
 
@@ -37,14 +38,16 @@ sealed class Expression(val range: Pair<Location, Location>) { // everything in 
         ) : Expression(range)
     }
 
-    class NestedExpression( // expression in parenthesis
+    // expression in parenthesis
+    class NestedExpression(
         range: Pair<Location, Location>,
         val expression: Expression,
     ) : Expression(range)
 
-    class SubsequentExpressions( // series of expressions separated by semicolons, returning value of the last one
-        // we agreed that (x; y;) returns Unit - should we create artificial class for empty expression after the last semicolon,
-        // or will we handle it in a different way?
+    // series of expressions separated by semicolons, returning value of the last one
+    // we agreed that (x; y;) returns Unit - should we create artificial class for empty expression after the last semicolon,
+    // or will we handle it in a different way?
+    class SubsequentExpressions(
         range: Pair<Location, Location>,
         vararg val expressions: Expression,
     ) : Expression(range)
--- a/src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
+++ b/src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
@@ -4,78 +4,43 @@ import cacophony.utils.Location
 
 sealed class Expression(val range: Pair<Location, Location>) { // everything in cacophony is an expression
     // artificial instance, can be useful when calculating values of nested expressions
-    class EmptyExpression(
+    class Empty(
         range: Pair<Location, Location>,
     ) : Expression(range)
 
     class Variable(
         range: Pair<Location, Location>,
+        val identifier: String,
     ) : Expression(range)
 
     // should Type also inherit expression? if so, can we do for example "; Bool;" in our language?
     class Type(
         range: Pair<Location, Location>,
+        val identifier: String,
     ) : Expression(range)
 
     class VariableDeclaration(
         range: Pair<Location, Location>,
-        val keywordLet: Keyword.KeywordLet,
+        val identifier: String,
     ) : Expression(range)
 
     class FunctionArgument(
         range: Pair<Location, Location>,
-        type: Type,
-    ) : Expression(range)
+        val type: Type,
+    )
 
     class VariableTypedDeclaration(
         range: Pair<Location, Location>,
-        variable: VariableDeclaration,
-        type: Type,
+        val variable: VariableDeclaration,
+        val type: Type,
     ) : Expression(range)
 
     class Literal(
         range: Pair<Location, Location>,
     ) : Expression(range)
 
-    class Keyword(
-        range: Pair<Location, Location>,
-    ) : Expression(range) {
-        class KeywordBreak(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-
-        // it is possible that following keywords are unnecessary, in this case break would be probably moved to statement
-        class KeywordReturn(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-
-        class KeywordLet(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-
-        class KeywordIf(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-
-        class KeywordThen(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-
-        class KeywordElse(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-
-        class KeywordWhile(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-
-        class KeywordDo(
-            range: Pair<Location, Location>,
-        ) : Expression(range)
-    }
-
     // expression in parenthesis
-    class NestedExpression(
+    class ParenthesisGroup(
         range: Pair<Location, Location>,
         val expression: Expression,
     ) : Expression(range)
@@ -84,7 +49,7 @@ sealed class Expression(val range: Pair<Location, Location>) { // everything in 
     // they can maybe be merged with NestedExpression, but I don't know if that would be convenient
     // we agreed that (x; y;) returns Unit - should we create artificial class for empty expression after the last semicolon,
     // or will we handle it in a different way?
-    class SubsequentExpressions(
+    class Subsequent(
         range: Pair<Location, Location>,
         vararg val expressions: Expression,
     ) : Expression(range)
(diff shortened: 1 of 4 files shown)


src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
+package cacophony.grammar.syntaxtree
+
+import cacophony.utils.Location
+
+sealed class Expression(val range: Pair<Location, Location>) { // everything in cacophony is an expression
+    class Variable(
+        range: Pair<Location, Location>,
+    ) : Expression(range)
+
+    // should Type also inherit expression? if so, can we do for example "; Bool;" in our language?
```
km1chno, line 10: good question, to discuss on Tuesday

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+    class Type(
+        range: Pair<Location, Location>,
+    ) : Expression(range)
+
+    class VariableDeclaration(
```
km1chno, line 15: shouldn't it consist of `let` keyword and variable identifier?

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+        range: Pair<Location, Location>,
+        variable: VariableDeclaration,
+        type: Type,
+    ) : Expression(range)
+
+    class Literal(
+        range: Pair<Location, Location>,
+    ) : Expression(range)
+
+    class Keyword(
```
km1chno, line 29: `let` seems to be missing

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+        val expression: Expression,
+    ) : Expression(range)
+
+    // series of expressions separated by semicolons, returning value of the last one
+    // we agreed that (x; y;) returns Unit - should we create artificial class for empty expression after the last semicolon,
+    // or will we handle it in a different way?
+    class SubsequentExpressions(
+        range: Pair<Location, Location>,
+        vararg val expressions: Expression,
+    ) : Expression(range)
```
km1chno, line 53: i wonder if we could merge it into one, as currently `NestedExpression` is just a special case of `SubsequentExpressions` for one expression

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    // expression in parenthesis
+    class NestedExpression(
+        range: Pair<Location, Location>,
+        val expression: Expression,
+    ) : Expression(range)
+
+    // series of expressions separated by semicolons, returning value of the last one
+    // we agreed that (x; y;) returns Unit - should we create artificial class for empty expression after the last semicolon,
```
km1chno, line 48: imo we should, it would just be an expression with empty range, based on that we could immediately deduce that its value is `Unit`

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+    // or will we handle it in a different way?
+    class SubsequentExpressions(
+        range: Pair<Location, Location>,
+        vararg val expressions: Expression,
+    ) : Expression(range)
+
+    sealed class Statement(
```
km1chno, line 55: shouldn't we also have `ReturnStatement` consisting of `ReturnKeyword` and expression?

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+            val testExpression: Expression,
+            val doExpression: Expression,
+            val elseExpression: Expression,
+        ) : Statement(range)
+
+        class WhileStatement(
+            range: Pair<Location, Location>,
+            val testExpression: Expression,
+            val doExpression: Expression,
+        ) : Statement(range)
```
km1chno, line 75: we could also add keywords `IfKeyword`, `ElseKeyword` and `WhileKeyword` and we could them here as children

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    class FunctionDefinition(
+        range: Pair<Location, Location>,
+        val functionName: VariableDeclaration,
+        vararg val arguments: VariableDeclaration,
```
km1chno, line 81: ok here I see why you needed `VariableDeclaration` to not have `let`

i think function arguments are something different than `VariableDeclaration`

km1chno: so they could have their own class

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,229 @@
...
+            class UnaryMinusOperator(
+                range: Pair<Location, Location>,
+                expression: Expression,
+            ) : UnaryOperator(range, expression)
+        }
+
+        sealed class BinaryOperator(
+            range: Pair<Location, Location>,
+            val expression1: Expression,
+            val expression2: Expression,
```
km1chno, line 112: all binary operators we have are infix, so i'd call these fields `lhs` and `rhs`

src/main/kotlin/cacophony/grammar/syntaxtree/Hierarchy.java
```diff
@@ -0,0 +1,25 @@
+package cacophony.grammar.syntaxtree;
+
+public class Hierarchy { //dont know how to implement it yet
```
km1chno, line 3: very cool you've done this work, I think all this will become clear after next classes on Tuesday

km1chno (changes requested):

src/main/kotlin/cacophony/grammar/syntaxtree/Hierarchy.java
```diff
@@ -0,0 +1,25 @@
+package cacophony.grammar.syntaxtree;
```
km1chno, line 1: please change file extensions from to `kt`

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: I think we do not need to keep let, if, while etc as separate class - from parent type it is already known what to do with children expression, so after parsing I think keeping them is unnecessary (also, 'let' by itself is not a valid expression)
However Im not sure, so if you are I can change it or we can discuss it on labs

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: As above

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: I will think about it for a moment if it wont mess anything with priority, also if it will be clear for later work

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: Sure, that should be done

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: Similarly to let

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: Thats right

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: Will change it

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: semantically they are the same, but I'm not sure if merging will be convenient for further implementation, so I will leave them separate for now

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,279 @@
...
+    ) : Expression(range)
+
+    class VariableTypedDeclaration(
+        range: Pair<Location, Location>,
+        variable: VariableDeclaration,
+        type: Type,
+    ) : Expression(range)
+
+    class Literal(
+        range: Pair<Location, Location>,
```
AHGIJMKLKKZNPJKQR, line 37: **suggestion:** As for `Variable` and `Type`, we could store the value of the literal. Also it might be good to distinguish between different types of literals here, i.e. something like `BooleanLiteral` (or even `True`, `False`), `IntegerLiteral`

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,279 @@
...
+            range: Pair<Location, Location>,
+        ) : Expression(range)
+
+        class KeywordDo(
+            range: Pair<Location, Location>,
+        ) : Expression(range)
+    }
+
+    // expression in parenthesis
+    class NestedExpression(
```
AHGIJMKLKKZNPJKQR, line 78: **suggestion:** maybe renaming this to `ParenthesisGroup` or would be more explicit (also `Expression.NestedExpression` repeats `Expression` )

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,279 @@
...
+            val testExpression: Expression,
+            val keywordDo: Keyword.KeywordDo,
+            val doExpression: Expression,
+        ) : Statement(range)
+
+        class ReturnStatement(
+            range: Pair<Location, Location>,
+            val keywordReturn: Keyword.KeywordReturn,
+            val value: Expression,
+        ) : Statement(range)
```
AHGIJMKLKKZNPJKQR, line 125: **issue:** I don't there's any point in storing the keywords for these node types, in general keywords useless after parsing, since they are only used in syntactic analysis, while the AST is for semantic analysis

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,279 @@
...
+        val functionName: VariableDeclaration,
+        vararg val arguments: FunctionArgument,
+        val returnType: Type,
+    ) : Expression(range)
+
+    class FunctionCall(
+        range: Pair<Location, Location>,
+        val functionName: Variable,
+        vararg val arguments: Variable,
+    ) : Expression(range)
```
AHGIJMKLKKZNPJKQR, line 139: **issue:** I don't think this works for scenarios like
`f[1][2]`, which should parse to something like
```
FunctionCall(FunctionCall("f", 1), 2)
```
I think `functionName` should be any `Expression`

AHGIJMKLKKZNPJKQR (changes requested):

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
AHGIJMKLKKZNPJKQR: I'm pretty sure there is no need to keep keywords in ast nodes. They are only useful when trying to parse an expression, once it is parsed there are redundant, all an `variable` declaration needs to know is the identifier of the variable and its type (and possibly value), there is no need to store `let`, `:` and `=`

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
AHGIJMKLKKZNPJKQR: I think `Type` should not be an expression, similarly `FunctionArgument` is not an expression

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
SebastianSpyrzewski: then maybe we should distinguish them at lexer level?

AHGIJMKLKKZNPJKQR: I think that would probably be the best

km1chno (approved):

AHGIJMKLKKZNPJKQR (approved):

PR #66, opened 2024-10-26
Find first

KacperTopolski:


src/main/kotlin/cacophony/grammars/Analysis.kt
```diff
@@ -1,44 +1,169 @@
...
+    }
+
+    private fun toState(entity: Entity<StateType, SymbolType, ResultType>): DFAStateReference<StateType, SymbolType, ResultType>? {
+        if (entity is Entity.Ref) {
+            return entity.ref
+        }
+        if (entity is Entity.Sym) {
+            return symbolToStateReference[entity.sym]
+        }
+        return null
```
km1chno, line 52: you can use pattern matching for such conditions

```kotlin
return when (entity) {
    is Entity.Ref -> entity.ref
    is Entity.Sym -> symbolToStateReference[entity.sym] 
}
```

src/main/kotlin/cacophony/grammars/Analysis.kt
```diff
@@ -1,44 +1,169 @@
...
+    }
+
+    private fun toSymbol(entity: Entity<StateType, SymbolType, ResultType>): SymbolType? {
```
km1chno, line 55: `fromState`, `fromSymbol`, `toState`, `toSymbol` should be methods of `Entity` class

src/main/kotlin/cacophony/grammars/Analysis.kt
```diff
@@ -1,44 +1,169 @@
...
 
-typealias DFAStateReference<StateType, AtomType, ResultType> = Pair<StateType, DFA<StateType, AtomType, ResultType>>
+typealias DFAStateReference<StateType, SymbolType, ResultType> =
+    Pair<StateType, DFA<StateType, SymbolType, ResultType>>
 
-fun <StateType, SymbolType> findNullable(
-    automata: Map<SymbolType, DFA<StateType, SymbolType, Production<SymbolType>>>,
-): Collection<DFAStateReference<StateType, SymbolType, Production<SymbolType>>> {
-    TODO("Not implemented")
+sealed class Entity<out Q, out A, out R> {
```
km1chno, line 9: `Entity` is such a generic name, please change it to something reflecting what is really is for

src/main/kotlin/cacophony/grammars/Analysis.kt
```diff
@@ -1,44 +1,169 @@
...
+): Collection<DFAStateReference<StateType, SymbolType, ResultType>> {
     TODO("Not implemented")
 }
 
-fun <StateType, SymbolType> findFollow(
-    automata: Map<SymbolType, DFA<StateType, SymbolType, Production<SymbolType>>>,
-    nullable: Collection<DFAStateReference<StateType, SymbolType, Production<SymbolType>>>,
-    first: Map<DFAStateReference<StateType, SymbolType, Production<SymbolType>>, Collection<SymbolType>>,
-): Map<DFAStateReference<StateType, SymbolType, Production<SymbolType>>, Collection<SymbolType>> {
+class FindFirstImpl<StateType, SymbolType, ResultType>(
```
km1chno, line 24: what's the advantage of creating this class instead of just putting the code in `findFirst()` function like it was declared before?

src/main/kotlin/cacophony/grammars/Analysis.kt
```diff
@@ -1,44 +1,169 @@
...
+        firstMap =
+            closure.mapNotNull { (key, values) ->
+                val newKey = toState(key) ?: return@mapNotNull null
+                val newValues = values.mapNotNull { toSymbol(it) }.toSet()
+                return@mapNotNull Pair(newKey, newValues)
+            }.toMap()
+    }
+}
+
+fun <VType> transitiveClosure(graph: Map<VType, Collection<VType>>): Map<VType, Set<VType>> {
```
km1chno, line 95: you can extract `transitiveClosure` and `findStronglyConnectedComponent` somewhere, they seem to be generic functions on graphs that could potentially be reused

and even if they won't be reused, you specifically made them generic and they just don't belong in this file

km1chno (changes requested):

src/main/kotlin/cacophony/grammars/Analysis.kt
KacperTopolski: fromSymbol and fromState are not member of Entity (or rather GeneralizedSymbol as it is now called) to allow them for implicit access to stateReferenceToSymbol and symbolToStateReference maps

src/main/kotlin/cacophony/grammars/Analysis.kt
KacperTopolski: Changed to GeneralizedSymbol

KacperTopolski:
Fundamentally, each "Entity" is either just a symbol (terminals), just a state (non starting states of automata) or both (non-terminals). I think storing entity in pair-like data class resembles this relation better, so I changed it.

src/main/kotlin/cacophony/grammars/Analysis.kt
KacperTopolski: Created GraphUtils in utils package. I also moved getReachableFrom there.

src/main/kotlin/cacophony/grammars/Analysis.kt
KacperTopolski: I had some problems to make it compile otherwise. Fixed now.

src/main/kotlin/cacophony/grammars/FindFirst.kt
```diff
@@ -1,10 +1,58 @@
...
+            graph[from]!!.add(by)
+            if (nullable.contains(by.state)) {
+                graph[from]!!.add(to)
+            }
+        }
+    }
+
+    return getTransitiveClosure(graph).mapNotNull { (key, values) ->
+        val newKey = key.state ?: return@mapNotNull null
+        val newValues = values.mapNotNull { it.symbol }.filter { !automata.containsKey(it) }.toSet()
```
KacperTopolski, line 55: TODO: The exact return value should be changed once a proper semantic is decided.

src/test/kotlin/cacophony/utils/GraphUtilsTest.kt
```diff
@@ -0,0 +1,70 @@
+package cacophony.utils
+
+import org.junit.jupiter.api.Assertions.assertEquals
+import org.junit.jupiter.api.Test
+
+class GraphUtilsTest {
```
km1chno, line 6: please add tests for `getReachableFrom` and `getStronglyConnectedComponents`

i know you are testing `getReachableFrom` in tests for closures but it is a public method so it should have its own test

src/main/kotlin/cacophony/grammars/FindFirst.kt
```diff
@@ -1,10 +1,62 @@
...
+
+data class GeneralizedSymbol<StateType, SymbolType, ResultType>(
+    val state: DFAStateReference<StateType, SymbolType, ResultType>?,
+    val symbol: SymbolType?,
+) {
+    init {
+        if (state == null && symbol == null) {
+            throw IllegalArgumentException("State and Symbol cannot both be nulls")
+        }
+    }
```
km1chno, line 14: you can set `state` and `symbol` to be non-nullable in class header and get rid of it

i see that if you do it, you get compilation error in two places below, but there you can do `!!` assertion and this way you achieve exactly the same as currently (if the reference actually turns out to be `null` in runtime, `KotlinNullPointerException` will be thrown)

mhorod:
> I'm approving but before merging, wait for https://github.com/mhorod/cacophony/pull/67 to be approved and verify that the tests there are passing, then we should merge https://github.com/mhorod/cacophony/pull/67 immediately after this PR.

We can also merge #67 directly to this PR to avoid splitting it over two PRs to main, I already changed the base to do that.

src/main/kotlin/cacophony/grammars/FindFirst.kt
KacperTopolski: I don't want them to be nulls at the same time. Either one of them can be null, as long as the other one is not.

PR #67, opened 2024-10-27
Add test for findFirst

mhorod:


km1chno (commented):
Very nice tests, I don't have any comments <3. But before I approve, let's wait for tomorrow to find out whether we should extend `first` sets with automata states

PR #74, opened 2024-10-29
Split `Literal` into `IntLiteral` and `BoolLiteral`

honvex-cool:
I also think that they could hold the actual Int/Boolean value, it will be needed sooner or later


AHGIJMKLKKZNPJKQR (commented):
`TokenCategorySpecific.LITERAL` should also be split

PR #76, opened 2024-10-29
Semantics interfaces

MisieqqeisiM:


src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,253 @@
...
+        range: Pair<Location, Location>,
+    ) {
+        class Basic(
+            range: Pair<Location, Location>,
+            val identifier: String,
+        ) : Type(range)
+
+        class Functional(
+            range: Pair<Location, Location>,
+            vararg val argumentsType: Type,
```
KacperTopolski, line 28: This will require a ton of asList() and spread operator calls in every function dealing with AST. Can't we just use a normal List?

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
```diff
@@ -0,0 +1,253 @@
...
+    sealed class Type(
+        range: Pair<Location, Location>,
+    ) {
+        class Basic(
+            range: Pair<Location, Location>,
+            val identifier: String,
+        ) : Type(range)
+
+        class Functional(
+            range: Pair<Location, Location>,
```
FilipKon13, line 27: Move the Type class outside of the Expression class (so that it is at the top level). Maybe make range a val field?
```Kotlin
sealed class Type(
    val range: Pair<Location, Location>,
) {
    class Basic(
        val identifier: String,
    ) : Type(range)

    class Functional(
        vararg val argumentsType: Type,
        val returnType: Type,
    ) : Type(range)
}
```

src/main/kotlin/cacophony/grammar/syntaxtree/Expression.kt
FilipKon13: I up that, we have similar interface in AlgebraicRegex and it was pain to cast everything to TypedArray every time. The vararg constructor is nice for testing, so I propose we add it as non-primary one.

FilipKon13:
Also, I propose to move all inner classes outside of Expression, so that we do not need to prefix every type with `Expression.`. Additionally, we can remove Operator and add UnaryOperator and BinaryOperator to avoid triple nesting (I do not really see a place when we want to use Operator specifically). After this changes, everything has at most one level of nesting, and logical division is still visible.

FilipKon13:
Look [here](https://github.com/mhorod/cacophony/tree/ast-types-proposal)

FilipKon13:
After the discussion with @SebastianSpyrzewski renamed the `Subsequent` class to `Block` ([here](https://github.com/mhorod/cacophony/tree/ast-types-proposal)) The whole AST is now represented by a `Block`

FilipKon13 (approved):

PR #87, opened 2024-11-01
Add value field to VariableDeclaration

MisieqqeisiM:


Rafi0221:
We should also add `body` to `FunctionDeclaration`

MisieqqeisiM:
> We should also add `body` to `FunctionDeclaration`

It has already been added by another PR

AHGIJMKLKKZNPJKQR (approved):

PR #89, opened 2024-11-01
Function call graph

AHGIJMKLKKZNPJKQR:


src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -42,21 +48,24 @@ sealed class Definition(
         range: Pair<Location, Location>,
         identifier: String,
         val type: Type.Basic?,
-    ) : Definition(range, identifier)
+    ) : Definition(range, identifier),
+        LeafExpression
```
Rafi0221, line 52: VariableDeclaration is not a leaf, see #87. Remember to recurse in it while generating call graph.

AHGIJMKLKKZNPJKQR: nice catch, I'll fix this when the other PR is merged

MisieqqeisiM (approved):

PR #90, opened 2024-11-01
Type checking

FilipKon13:
Initial version to allow an early feedback - no diagnostics atm, the tests do not cover checking errors


src/test/kotlin/cacophony/semantic/TypeCheckerTest.kt
```diff
@@ -0,0 +1,952 @@
...
+    private val intLiteral = Literal.IntLiteral(lc, 7)
+    private val booleanLiteral = Literal.BoolLiteral(lc, true)
+
+    @Test
+    fun `ok - empty block`() {
+        val ast = Block(lc, emptyList())
+        val diagnostics = getDiagnostic()
+        val result = checkTypes(ast, diagnostics, emptyMap())
+        assertTypeEquals(BuiltinType.UnitType, result[ast])
+        assertNull(diagnostics.msg)
```
km1chno, line 72: you can do

```
@MockK
lateinit var diagnostics: Diagnostics

@BeforeEach
fun setUpMocks() {
    MockKAnnotations.init(this, relaxUnitFun = true)
    every { diagnostics.report(any(), any<Location>()) } just runs
}
```

and then in tests

```
verify(exactly = 0) { diagnostics.report(any(), any<Location>()) }
```

like in https://github.com/mhorod/cacophony/blob/53493d26df99c0c86e1fb7beedbe9c2536f554aa/src/test/kotlin/cacophony/lexer/RegularLanguageLexerTest.kt#L98

FilipKon13: I can, but should I?

km1chno: up to you, I prefer it this way but its a matter of taste

src/main/kotlin/cacophony/semantic/TypeChecker.kt
```diff
@@ -1,19 +1,345 @@
...
+                is OperatorUnary.Negation -> {
+                    val innerType = typeExpression(expression.expression) ?: return null
+                    if (!isSubtype(innerType, BuiltinType.BooleanType)) {
+                        diagnostics.report("Type $innerType does not support unary ! operator", expression.expression.range)
+                        return null
+                    }
+                    BuiltinType.BooleanType
+                }
+                // TODO: who should check if break is inside loop?
+                is Statement.BreakStatement -> BuiltinType.UnitType
```
MisieqqeisiM, line 201: break should have Void type

FilipKon13:
There are some TODOs refering to code duplication - I intend to remove those before the merge. Additionally, there is inconsistency in how name of function type is generated, but it does not show up atm because we do not have function as args - I will fix this as well

src/main/kotlin/cacophony/semantic/TypeChecker.kt
FilipKon13: I will add test for this as well

MisieqqeisiM (approved):
I like it

PR #93, opened 2024-11-02
Add grammar for cacophony

jjaworska:


src/main/kotlin/cacophony/semantic/ASTGeneration.kt
```diff
@@ -0,0 +1,327 @@
...
+                    throw IllegalArgumentException("Expected the operator symbol, got: $operator")
+                }
+            }
+            ADDITION_LEVEL, MULTIPLICATION_LEVEL, EQUALITY_LEVEL, COMPARATOR_LEVEL, LOGICAL_OPERATOR_LEVEL -> {
+                assert(childNum >= 3)
+                operatorRegexToAST(parseTree.children, diagnostics)
+            }
+            UNARY_LEVEL -> {
+                assert(childNum == 2)
+                val operator = parseTree.children[0]
```
MisieqqeisiM, line 289: Please refrain from using Kotlin keywords as variable names.

src/main/kotlin/cacophony/parser/CacophonyGrammarSymbol.kt
```diff
@@ -1,70 +1,97 @@
+@file:Suppress("ktlint:standard:no-wildcard-imports")
```
MisieqqeisiM, line 1: This annotation will be unnecessary after merging, since we have enabled wildcard imports in the whole project.

MisieqqeisiM (approved):

PR #100, opened 2024-11-04
AST small tests and improved exception handling

KacperTopolski:


KacperTopolski:
I tried to somehow standardize exception types, so I propose the following: change all *ErrorException classes to either *Exception, representing normal compilation errors, or *Error, representing error of compiler itself. Each *Exception class is derived from CompileException.

src/main/kotlin/cacophony/parser/CacophonyGrammarSymbol.kt
```diff
@@ -1,3 +1,5 @@
+@file:Suppress("ktlint:standard:no-wildcard-imports")
```
MisieqqeisiM, line 1: This is unnecessary

KacperTopolski: This is already on the main branch :)

MisieqqeisiM (approved):

PR #105, opened 2024-11-05
Add interfaces for control flow

Rafi0221:


src/main/kotlin/cacophony/controlflow/CFGNode.kt
```diff
@@ -0,0 +1,9 @@
+package cacophony.controlflow
+
+sealed interface CFGRootNode
+
+sealed interface CFGLeafNode
+
+sealed class CFGNode {
+    // TODO
+}
```
AHGIJMKLKKZNPJKQR, line 9: **suggestion:** In my PR I opted for a `CFGVertex` class that contains a `CFGNode` to represent a vertex with its neighbors. I think you can just rename `CFGRootNode` to `CFGVertex` here and I can then resolve any conflicts once this is merged (we can also go for `CFGRootNode` but I think `CFGVertex` is less confusing.

src/main/kotlin/cacophony/controlflow/Variable.kt
```diff
@@ -0,0 +1,17 @@
...
+    val definition: Definition,
+) : Variable()
+
+sealed class Register() : Variable()
+
+data object VirtualRegister : Register()
+
+data object FixedRegister : Register() {
+    // TODO
+}
```
AHGIJMKLKKZNPJKQR, line 17: **question:** Why are these singleton objects and not classes?

src/main/kotlin/cacophony/controlflow/CFGNode.kt
Rafi0221: I agree. I just wanted some superclass for all roots and leaves. I will also delete `CFGLeafNode` as you have `Leaf` in your PR.

src/main/kotlin/cacophony/controlflow/Variable.kt
Rafi0221: Right, they should be classes.

src/main/kotlin/cacophony/controlflow/Variable.kt
```diff
@@ -0,0 +1,17 @@
...
+
+sealed class Variable
+
+class SourceVariable(
+    val definition: Definition,
+) : Variable()
+
+sealed class Register() : Variable()
+
+class VirtualRegister : Register()
```
Rafi0221, line 13: **Question to all:** Do we want to leave `VirtualRegister` outside of `Register` or maybe put it inside as `Register.Virtual`?

src/main/kotlin/cacophony/controlflow/Variable.kt
```diff
@@ -0,0 +1,17 @@
...
+
+import cacophony.semantic.syntaxtree.Definition
+
+sealed class Variable
+
+class SourceVariable(
+    val definition: Definition,
+) : Variable()
+
+sealed class Register() : Variable() {
```
AHGIJMKLKKZNPJKQR, line 11: **suggestion:** 
```suggestion
sealed interface Variable

class SourceVariable(
    val definition: Definition,
) : Variable

sealed class Register() : Variable {
```

src/main/kotlin/cacophony/controlflow/Variable.kt
AHGIJMKLKKZNPJKQR, line 1: **issue:** We are missing a type for internal variables that are stack-only, since we can't use `SourceVariable` for them as they have no corresponding `Definition`

AHGIJMKLKKZNPJKQR (approved):

PR #119, opened 2024-11-09
Remove short-circuiting operators from CFGNode

AHGIJMKLKKZNPJKQR:
These should not be used to allow implementing of short-circuiting


honvex-cool (approved):
Thanks for removing them :D

PR #121, opened 2024-11-09
Simple analysis of variable use types

SebastianSpyrzewski:
Added analysis of variable use types - for each expression there is a map of variables used by this expression and types of those uses (read, write, read/write) - analysis is similar to FunctionAnalysis

Questions:
1. For now expression marks variable as "used" even if it is declared inside a nested expression or it is an argument of function declared inside it - for example
```
let f = [] -> Unit =>(
  let h = [x:Int] -> Unit =>(
    let y: Int = 2;
  )
)
```
function f will have both x and y marked as used - should we get rid of it?

2.  There will be a lot of maps for long programs, and in my understanding we need them only for certain expressions - maybe we can optimize it to not return a map including every existing expression?

3. I did not use a CallGraph argument, is that fine?


src/main/kotlin/cacophony/semantic/VarUseTypeAnalysis.kt
```diff
@@ -11,5 +9,227 @@ fun analyzeVarUseTypes(
...
+            is FunctionCall -> visitFunctionCall(expr)
+            is Statement.IfElseStatement -> visitIfElseStatement(expr)
+            is Statement.WhileStatement -> visitWhileStatement(expr)
+            is Statement.ReturnStatement -> visitReturnStatement(expr)
+            is OperatorUnary -> visitUnaryOperator(expr)
+            is OperatorBinary -> visitBinaryOperator(expr)
+            is VariableUse -> visitVariableUse(expr)
+            else -> {
+                // do nothing for expressions without nested expressions
+            }
```
Rafi0221, line 78: You can use `LeafExpression` here

src/main/kotlin/cacophony/semantic/VarUseTypeAnalysis.kt
```diff
@@ -11,5 +9,227 @@ fun analyzeVarUseTypes(
...
+            expr,
+            UseTypesForExpression.merge(
+                useTypeAnalysis.get(expr.testExpression),
+                useTypeAnalysis.get(expr.doExpression),
+                useTypeAnalysis.get(expr.elseExpression),
+            ),
+        )
+    }
+
+    private fun visitFunctionCall(expr: FunctionCall) {
```
Rafi0221, line 191: I think here you have to check how this function uses variables (probably via `analyzedFunction`, so ad. 3. you might need this, instead of `callGraph`, sorry). e.g. in `x+f[]`, where `f` uses `x` , `x` should be marked as used in `f[]`, so later the control flow is correct.

Rafi0221 (dismissed):
General idea seems to be fine. Please add tests. Also, IntelliJ has a lot of weak warnings (like using put and get...).

Rafi0221:
Ad 1. and 2. I think we only need to keep track of variables that are visible in current scope

src/main/kotlin/cacophony/semantic/StaticFunctionRelations.kt
```diff
@@ -118,6 +124,15 @@ private class StaticFunctionsRelationsVisitor {
             is VariableUse -> visitVariableWrite(expr.lhs)
             else -> visitExpression(expr.lhs)
         }
+        visitExpression(expr.rhs)
+    }
+
+    private fun visitCompoundAssignment(expr: OperatorBinary) {
```
mhorod, line 130: nice that you spotted that :D

src/main/kotlin/cacophony/semantic/VarUseTypeAnalysis.kt
```diff
@@ -1,15 +1,243 @@
...
+        }
+    }
+}
+
+private class VarUseVisitor(
+    val resolvedVariables: ResolvedVariables,
+) {
+    private val useTypeAnalysis = mutableMapOf<Expression, UseTypesForExpression>()
+    private val scopeStack = ArrayDeque<MutableSet<Definition>>()
+    var staticDepth = -1
```
mhorod, line 62: this var is no longer used

src/main/kotlin/cacophony/semantic/VarUseTypeAnalysis.kt
```diff
@@ -1,15 +1,243 @@
...
+            visitExpression(it)
+            useTypeAnalysis[expr]!!.mergeWith(useTypeAnalysis[it]!!)
+        }
+
+        // variables used in called function:
+        val calledFunction = resolvedVariables[expr.function]
+        if (calledFunction is Definition.FunctionDeclaration) {
+            val map =
+                useTypeAnalysis[calledFunction.body]!!.getMap().filterNot {
+                    // had to use staticDepth because of called functions that use variables not visible in current scope
```
mhorod, line 208: staticDepth seems to not be used anymore

src/main/kotlin/cacophony/semantic/VarUseTypeAnalysis.kt
```diff
@@ -1,15 +1,243 @@
...
+                    it.key in calledFunction.arguments
+                }
+            useTypeAnalysis[expr]!!.mergeWith(UseTypesForExpression(map.toMutableMap()))
+        } else {
+            throw IllegalStateException("Left side of function call is not a function declaration")
```
mhorod, line 213: `throw IllegalStateException` can be replaced with `error`

src/main/kotlin/cacophony/semantic/VarUseTypeAnalysis.kt
```diff
@@ -1,15 +1,243 @@
...
+    fun getAnalysisResult(): UseTypeAnalysisResult = useTypeAnalysis.mapValues { it.value.getMap() }
+
+    private fun visitExpression(expr: Expression) {
+        when (expr) {
+            is Definition -> {
+                scopeStack.lastOrNull()?.add(expr)
+            }
+            else -> {
+            }
+        }
```
mhorod, line 75: we can replace this with just `if (expr is Definition)`

src/main/kotlin/cacophony/semantic/VarUseTypeAnalysis.kt
```diff
@@ -1,15 +1,243 @@
...
+        expr.arguments.forEach { // arguments should be marked as read-used
+            visitExpression(it)
+            useTypeAnalysis[expr]!!.mergeWith(useTypeAnalysis[it]!!)
+        }
+
+        // variables used in called function:
+        val calledFunction = resolvedVariables[expr.function]
+        if (calledFunction is Definition.FunctionDeclaration) {
+            val map =
+                useTypeAnalysis[calledFunction.body]!!.getMap().filterNot {
```
mhorod, line 207: How does this work with recursion, and in particular with mutual recursion?
I'm worried that this line will throw an exception due to body not being defined.

It seems this is the place we should use `functionAnalysis`, no?
We can look at `functionAnalysis[calledFunction].variables` and check which variables are declared outside of it (i.e. where `definedIn != calledFunction` since this means that variable comes from parent we are in)

mhorod (changes requested):

src/test/kotlin/cacophony/semantic/FunctionAnalysisTest.kt
```diff
@@ -30,7 +30,7 @@ class FunctionAnalysisTest {
             )
     }
 
-    @Test
```
mhorod, line 33: Why are these lines commented out?

SebastianSpyrzewski: These tests don't make any sense now as function don't keep info about usage of variables created by itself.
(for now every test return just returns function with empty variables set)
However, when the whole AST will be the function, these tests will be adjusted appropriately, so I thought there is no need to remove them completely just to write almost identical tests in a moment.

mhorod:
LGTM 👍 
Just fix the conflicts and we can merge

mhorod (approved):

Issue #114, opened 2024-11-05
Static links parameters in FunctionHandler


Rafi0221:
Implement `introduceStaticLinksParams` producting extra parameters we need to add to the function calls that are needed for the static links to work.

PR #122, opened 2024-11-09
Add generate variable access

### LINKED ISSUE #112: Generate variable access in FunctionHandler
Implement `generateVariableAccess` using static links.

km1chno:
Resolves #112 

This is an initial version for early feedback

## Constroversial changes
- Added `ancestorFunctionHandlers` to constructor of `FunctionHandler`
- Removed the `framePointer` parameter, because I think it's just `rbp`?

## TODO
- [x] tests
- [x] using static params which are supposed to be introduced as part of #114


src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -1,37 +1,118 @@
...
+    ) : VariableAllocation()
 
-        data class Stack(val offset: Int)
-    }
+    data class Stack(
+        val offset: Int,
+    ) : VariableAllocation()
+}
+
+interface FunctionHandler {
```
FilipKon13, line 24: Why would we change that to an interface? Do we expept another implementation or are there any problems with mocking for the tests?

km1chno: yes, its for tests
can you easily mock non-interfaces in kotlin?

Rafi0221 (changes requested):
Looks good, usage of `getVariableAllocation` alligns well with Filip's task. Please implement tests 😺

Rafi0221 (approved):

PR #123, opened 2024-11-09
Generate call

KacperTopolski:


src/main/kotlin/cacophony/controlflow/CFGNode.kt
```diff
@@ -40,8 +40,9 @@ sealed interface CFGNode {
         Unconditional,
         Leaf
 
-    class Call :
-        Unconditional,
+    class Call(
+        val identifier: String,
```
mhorod, line 44: shouldn't it be `FunctionDeclaration` ?

src/main/kotlin/cacophony/controlflow/X64Register.kt
```diff
@@ -0,0 +1,30 @@
+package cacophony.controlflow
+
+enum class X64Register(val isCallPreserved: Boolean) {
```
km1chno, line 3: maybe `Register.Fixed` could be parametrized by such enum, in #124 I added `id: String` but it is a temporary solution

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -40,8 +41,82 @@ class FunctionHandlerImpl(
...
+                        ),
+                    ),
+                ),
+            )
+        }
+
+        // in what order should we evaluate arguments? gcc uses reversed order
+        for ((argument, register) in registerArguments) {
+            nodes.add(CFGNode.Assignment(Register.FixedRegister(register), argument))
+        }
```
Giselus, line 83: Shouldn't we use getVariableAllocation to see where variables go?

KacperTopolski: No, as this is a part of a calling convention.

src/main/kotlin/cacophony/controlflow/X64Register.kt
KacperTopolski: I think there is no need for that, as we are most probably going to stick to x86_64 architecture.

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -40,8 +41,82 @@ class FunctionHandlerImpl(
...
+        val nodes: MutableList<CFGNode> = mutableListOf()
+
+        if (respectStackAlignment) {
+            // we push two copies of RSP to the stack and either leave them both there,
+            // or remove one of them via RSP assignment
+            val oldRSP = Register.VirtualRegister()
+            nodes.add(CFGNode.Assignment(oldRSP, CFGNode.VariableUse(Register.FixedRegister(X64Register.RSP))))
+            for (i in 0..2) {
+                nodes.add(CFGNode.Push(CFGNode.VariableUse(oldRSP)))
+            }
```
Rafi0221, line 58: This pushes 3 times, because range expression is close-ended by default.

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -40,8 +41,82 @@ class FunctionHandlerImpl(
...
+                    ),
+                ),
+            )
+        }
+
+        if (respectStackAlignment) {
+            // we could remove the operations from previous `if (stackArguments.isNotEmpty())` block
+            // via MemoryAccess, but for now the semantics are a bit unclear + it would introduce
+            // a few ifs, which we do not need at this point
+            nodes.add(CFGNode.Pop(Register.FixedRegister(X64Register.RSP)))
```
Rafi0221, line 111: I'm not sure if I understand, but you pushed RSP 2 times, but here you pop it only once.

KacperTopolski: I pop it directly to RSP, which implicitly pops the second RSP entry if it is there.

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
KacperTopolski: Fixed. I unrolled the loop to make the code clearer.

src/main/kotlin/cacophony/controlflow/X64Register.kt
Rafi0221: I mean, we can put this enum in `Register.Fixed` instead of `String`

Rafi0221 (changes requested):
Please add tests

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -40,8 +41,82 @@ class FunctionHandlerImpl(
...
+        }
+
+        if (result != null) {
+            nodes.add(CFGNode.Assignment(result, CFGNode.VariableUse(Register.FixedRegister(X64Register.RAX))))
+        }
+
+        // should this be a sequence?
+        return mapOf(CFGLabel() to CFGVertex.Final(CFGNode.Sequence(nodes)))
```
Rafi0221, line 119: To align with Jan's PR, yes, this should be a sequence

src/main/kotlin/cacophony/controlflow/X64Register.kt
KacperTopolski: Done

KacperTopolski:
> Please add tests

Done

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
KacperTopolski: Jan's PR requires this to be a List<CFGNode>. I changed my PR to match.

Rafi0221 (approved):

PR #125, opened 2024-11-09
Extend tests

MisieqqeisiM:


src/test/kotlin/cacophony/semantic/TypeCheckerTest.kt
```diff
@@ -1014,11 +1026,11 @@ class TypeCheckerTest {
     fun `error - function declaration type mismatch - types`() {
         val arg1 = Definition.FunctionArgument(lc, "x", testInt)
         val arg2 = Definition.FunctionArgument(lc, "y", testBoolean)
-        val funDef = Definition.FunctionDeclaration(lc, "f", Type.Functional(lc, listOf(Type.Basic(lc, "Boolean"), Type.Basic(lc, "Int")), Type.Basic(lc, "Int")), listOf(arg1, arg2), testInt, intLiteral)
+        val funDef = Definition.FunctionDeclaration(lc, "f", Type.Functional(lc, listOf(testBoolean, testInt), testInt), listOf(arg1, arg2), testInt, intLiteral)
```
FilipKon13: I intentionally created here new instances, co that reference-wise equality semantic is not violated (this would not be the same instance if it came from the code)

examples/correct/functions_basic.cac
```diff
@@ -2,14 +2,13 @@ let simple_check = [x: Int] -> Bool => x == 42; # the body of a function can be
...
 let complex_check = [x: Int] -> Bool => (
-    let temporary = simple[x];
+    let temporary = simple_check[x];
     temporary || x == 43
 ); # the body of a function can also be a block
 complex_check[43]; # returns `true`
 
-let global = false;
 let confusing = [x: Int] -> Bool =>
-    global = true; # careful! this is syntactically correct but the indentation is confusing
```
Rafi0221, line 12: Why did you remove global variable? Maybe it is not compiling right now (because we don't have static links to global scope, but it should compile in the future).

src/main/kotlin/cacophony/semantic/TypeChecker.kt
```diff
@@ -347,41 +358,44 @@ private class ErrorHandler(val diagnostics: Diagnostics) {
...
+sealed class BuiltinType private constructor(
+    name: String,
+) : TypeExpr(name) {
+    object BooleanType : BuiltinType("Bool")
 
     object IntegerType : BuiltinType("Int")
 
     object UnitType : BuiltinType("Unit")
+
+    object VoidType : BuiltinType("Void")
```
Rafi0221, line 384: Moving Void to `BuiltinType` causes problems with subtyping in function arguments and return type. So, to make it properly work, we should discuss how subtyping interacts with overloading (and I think we decided to not do that right now). I propose to remove Void from `BuiltinType` for now.

src/main/kotlin/cacophony/diagnostics/Diagnostics.kt
```diff
@@ -0,0 +1,15 @@
+package cacophony.diagnostics
+
+import cacophony.utils.Location
+
+interface Diagnostics {
+    fun report(
+        message: DiagnosticMessage,
+        location: Location,
+    ) = report(message, location to location)
```
Rafi0221, line 9: Input a nicer `toString` on a single `Location`. Now, we wouldn't use it. So maybe change `locationRangeToString` to call `locationToString` given range of length 1.

Rafi0221 (changes requested):

examples/correct/functions_basic.cac
MisieqqeisiM: I will move this part of code to "not yet correct" folder, since we won't be supporting global variables anytime soon

src/test/kotlin/cacophony/semantic/TypeCheckerTest.kt
MisieqqeisiM: Sorry, now it should be fixed

src/main/kotlin/cacophony/semantic/TypeChecker.kt
MisieqqeisiM: Done

Rafi0221 (approved):

Rafi0221:
Please add `syncSymbols` in `CacophonyParser.kt`

PR #126, opened 2024-11-10
Introduce static links

Giselus:
I don't think it's even close to being ready, there is a lot of things to discuss (and I haven't created any tests 😿).

1)I think previous interface was missing few things, to properly generate and pass this static links we also needed to know caller function, so I slightly modified it, maybe in the wrong way.

2)On the lecture it seems like every function call gets additional argument which is staticLink to parent frame. Let's suppose we have outer function, that has inner function inside. 

- If we call inner from inner for example everything is alright, we can pass outer_link, because we have it. 
- If we call outer from inner everyhing is alright too, even if outer needs some other link because example is more nested, link will be resolved as variable.
- But what if we want to call inner from outer, we would have to pass outer_link, so to make it easy it would be nice if outer function has outer_link (if it doesn't we have to make a little if, which isn't that bad).

So right now function stores both link to itself and to parent. It makes Katzper's task easier as well I think, because when he finds function that has the variable we want to use, he can ask about link to this function directly, instead of asking from it's child, but it's small problem and it could be solved without it anyway.

One thing I'm not so sure about is that now two functions have exactly the same variable in auxVariables, I don't know if it can lead to any problems or if it's completely normal thing.

Also maybe something more is wrong, because I may have misunderstood something about variable creation and stuff, maybe Filip's PR will shine some light onto it later.

### DIFF
--- a/src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
+++ b/src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
@@ -1,5 +1,9 @@
 package cacophony.controlflow
 
+import cacophony.controlflow.CFGNode.Addition
+import cacophony.controlflow.CFGNode.Constant
+import cacophony.controlflow.CFGNode.MemoryAccess
+import cacophony.controlflow.CFGNode.VariableUse
 import cacophony.semantic.AnalyzedFunction
 import cacophony.semantic.syntaxtree.Definition
 import cacophony.semantic.syntaxtree.Definition.FunctionDeclaration
@@ -35,6 +39,11 @@ interface FunctionHandler {
 
     fun getVariableAllocation(variable: Variable): VariableAllocation
 
+    fun registerVariableAllocation(
+        variable: Variable,
+        allocation: VariableAllocation,
+    )
+
     // Returns static link to parent
     fun getStaticLink(): Variable.AuxVariable.StaticLinkVariable
 
@@ -52,8 +61,8 @@ class FunctionHandlerImpl(
     private val ancestorFunctionHandlers: List<FunctionHandler>,
 ) : FunctionHandler {
     private val staticLink: Variable.AuxVariable.StaticLinkVariable = Variable.AuxVariable.StaticLinkVariable()
-
     private val definitionToVariable = analyzedFunction.variables.associate { it.declaration to Variable.SourceVariable(it.declaration) }
+
     private val variableAllocation: MutableMap<Variable, VariableAllocation> =
         run {
             val res = mutableMapOf<Variable, VariableAllocation>()
@@ -78,7 +87,7 @@ class FunctionHandlerImpl(
         introduceStaticLinksParams()
     }
 
-    private fun registerVariableAllocation(
+    override fun registerVariableAllocation(
         variable: Variable,
         allocation: VariableAllocation,
     ) {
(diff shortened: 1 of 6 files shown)


src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -52,7 +99,15 @@ class FunctionHandlerImpl(
...
     }
 
-    private fun introduceStaticLinksParams(): List<CFGNode> {
-        TODO("Not yet implemented")
+    override fun getStaticLink(): Variable.AuxVariable.StaticLinkVariable {
+        return staticLink
+    }
+
+    // Creates staticLink auxVariable in analyzedFunction, therefore shouldn't be called multiple times.
+    private fun introduceStaticLinksParams() {
```
Rafi0221, line 107: Maybe you can call it in the init block of `FunctionHandler`?

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -20,11 +20,19 @@ interface FunctionHandler {
     fun generateCall(
         arguments: List<CFGNode>,
         result: Register?,
+    ): List<CFGNode>
+
+    fun generateCallFrom(
+        function: FunctionHandler,
```
km1chno, line 26: argument name should be consistent with function in implementation

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -34,16 +42,55 @@ class GenerateVariableAccessException(
...
+        result: Register?,
+    ): CFGFragment {
+        val nodes: MutableList<CFGNode> = mutableListOf()
+        if (ancestorFunctionHandlers.isNotEmpty()) {
+            val parentFunction = ancestorFunctionHandlers[0]
+            val parentLink = parentFunction.getStaticLink()
+            nodes.add(
+                CFGNode.MemoryWrite(
+                    CFGNode.MemoryAccess(caller.generateVariableAccess(parentLink)),
+                    CFGNode.MemoryAccess(this.generateVariableAccess(parentLink)),
```
km1chno, line 79: I don't understand the calling convention fully, so this is a sanity check: does it work if `caller === function`? So in case of recursion?

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -52,7 +99,15 @@ class FunctionHandlerImpl(
...
 
-    private fun introduceStaticLinksParams(): List<CFGNode> {
-        TODO("Not yet implemented")
+    override fun getStaticLink(): Variable.AuxVariable.StaticLinkVariable {
+        return staticLink
+    }
+
+    // Creates staticLink auxVariable in analyzedFunction, therefore shouldn't be called multiple times.
+    private fun introduceStaticLinksParams() {
+        analyzedFunction.auxVariables.add(staticLink)
```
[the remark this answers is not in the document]
km1chno: I answered my own last question: I just call `getStaticLink()` in according function handler, right?

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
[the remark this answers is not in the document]
Giselus: generateVariableAccess returns only CFGNode, I don't know if it's register and I don't even know if I'm the one who should care about this.

Giselus: If you try to generateVariableAccess to your own link now, I guess you'll should check that this variable is on the variable list of this function, so you have it and you can obtain its allocation.

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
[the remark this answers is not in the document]
Giselus: Now it's only one, so it's less of a problem. But it's staticLink to parent function, so if you want to get staticLink to f, you should do g.getStaticLink(), where g is first function on a path towards variable access.

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -52,7 +102,19 @@ class FunctionHandlerImpl(
...
+    }
+
+    // Creates staticLink auxVariable in analyzedFunction, therefore shouldn't be called multiple times.
+    // Static link is created even if parent doesn't exist.
+    private fun introduceStaticLinksParams() {
+        // I truly don't know who is responsible for layout of the stack
+        // TODO: uncomment (and fix) after #127 is merged
+//            registerVariable(
+//                staticLink,
+//                VariableAllocation.OnStack(/*what should be there?*/),
```
Rafi0221, line 116: I think you can allocate it in a register, later register coloring will likely put it on stack

Rafi0221 (changes requested):
This might be fine, but please add tests 😺

PR #130, opened 2024-11-12
Add CFG generation

honvex-cool:
Add generation of CFG with vertices containing simple operations expressed as nodes


jjaworska (approved):
LGTM

PR #138, opened 2024-11-16
CFG tests

mhorod:


src/main/kotlin/cacophony/semantic/FunctionAnalysis.kt
```diff
@@ -14,13 +14,28 @@ data class ParentLink(
...
     val parentLink: ParentLink?,
     val variables: Set<AnalyzedVariable>,
     val auxVariables: MutableSet<Variable.AuxVariable>,
     val staticDepth: Int,
     val variablesUsedInNestedFunctions: Set<Definition>,
-)
+) {
+    fun declaredVariables() = variables.filter { it.definedIn == function }
+
+    fun outerVariables() = variables.filter { it.definedIn != function }
```
honvex-cool, line 37: Maybe these could be `val ... by lazy {...}` to compute it just once and only if necessary?

honvex-cool (changes requested):
Fantastic framework for testing :D
Please apply the discussed cleanup and resolve conflicts, and we will be able to merge :)

src/main/kotlin/cacophony/semantic/FunctionAnalysis.kt
mhorod: snapshots don't compile with such solution ☹️

honvex-cool (approved):
LGTM :+1:

PR #139, opened 2024-11-16
Change AST to function

SebastianSpyrzewski:


__snapshot__/correct/blocks_hard.cac.resolved.snap
```diff
@@ -1,3 +1,14 @@
 {
-  "__class__": "EmptyMap"
+  "__class__": "SingletonMap",
+  "(Location(value\u003d90), Location(value\u003d90))": {
```
KacperTopolski, line 3: These "value\u003d90" look quite strange, is it intended?

SebastianSpyrzewski: This value was displayed the same way on previous snapshots, so I thought it is OK

src/main/kotlin/cacophony/semantic/ASTGeneration.kt
```diff
@@ -292,11 +292,48 @@ private fun generateASTInternal(
     return Empty(range = Pair(Location(0), Location(0)))
 }
 
+private fun wrapInFunction(originalAST: AST): AST {
+    val beforeStart = Location(-1)
+    val behindEnd = Location(originalAST.range.second.value + 1)
+    val program =
+        Definition.FunctionDeclaration(
+            Pair(beforeStart, behindEnd),
+            "<program>",
```
honvex-cool, line 301: The string `"<program>"` occurs a few times in the code. I'm wondering whether we should create `const val PROGRAM_FUNCTION_IDENTIFIER = "<program>"` or `const val MAIN_FUNCTION_IDENTIFIER = "<program>"`. However, for now this might be an overkill, so this is just a suggestion, the decision is yours.

src/main/kotlin/cacophony/semantic/ASTGeneration.kt
```diff
@@ -292,11 +292,48 @@ private fun generateASTInternal(
...
+
 fun generateAST(
     parseTree: ParseTree<CacophonyGrammarSymbol>,
     diagnostics: Diagnostics,
 ): AST {
     val prunedTree = pruneParseTree(parseTree, diagnostics)!!
-    val ast = generateASTInternal(prunedTree, diagnostics) as AST
-    return ast
+    val ast = generateASTInternal(prunedTree, diagnostics)
+    return wrapInFunction(ast)
```
honvex-cool, line 338: Now that the whole program is wrapped in a main function, toplevel `return ()` is valid. Could you please add a minimalistic test to document this behavior too? (Sorry if there is one already, I must have missed it)

honvex-cool (changes requested):
Great work, thank you for contributing :D
I just left some minor comments, but these are not strong opinions, so feel free to mark the discussions as resolved even if you don't introduce any more changes.

honvex-cool (approved):
LGTM :+1:

PR #141, opened 2024-11-18
Basic patterns

### LINKED ISSUE #132: Prepare a basic set of patterns
Use the `Pattern` class hierarchy to create a set of patterns that will enable us to capture the behavior of all possible trees in the CFG. These patterns will map to a basic instruction set in the future, but for now, no specific interface is provided for `Instruction`, so feel free to add whatever you deem necessary.

jjaworska:
Resolves #132 (at least partially)


src/main/kotlin/cacophony/codegen/patterns/CacophonyPattern.kt
```diff
@@ -1,22 +1,215 @@
...
+    abstract class UnaryOpPattern {
+        val childLabel = ValueLabel()
+    }
+
+    // TODO: reduce boilerplate (typed DSL?)
+
+    // for now, we can always dump a constant into a register and call it a value
+    object ConstantPattern : ValuePattern {
+        val label = ConstantLabel()
+        val predicate: (Int) -> Boolean = { _ -> false }
```
mhorod, line 25: shouldn't this be `_ -> true` so that all constants are catched? 🤔

src/main/kotlin/cacophony/codegen/patterns/CacophonyPattern.kt
```diff
@@ -27,4 +220,223 @@ class CacophonyPattern {
...
+        override fun makeInstance(fill: SlotFill): List<Instruction> {
+            TODO("Not yet implemented")
+        }
+    }
+
+    object PopPattern : SideEffectPattern {
+        val regLabel = RegisterLabel()
+
+        override val tree = CFGNode.Pop(CFGNode.RegisterSlot(regLabel))
+//        override val tree = CFGNode.Pop()
```
mhorod, line 412: NIT: remove this commented line

src/main/kotlin/cacophony/codegen/patterns/Pattern.kt
```diff
@@ -7,7 +7,7 @@ import cacophony.controlflow.*
 data class SlotFill(
     val valueFill: Map<ValueLabel, CFGNode.Value>,
```
mhorod, line 8: I think this should be `Map<ValueLabel, VirtualRegister>`, but we can also talk about this on tommorow labs

src/main/kotlin/cacophony/codegen/patterns/CacophonyPattern.kt
```diff
@@ -1,22 +1,215 @@
...
 
 import cacophony.codegen.instructions.Instruction
-import cacophony.controlflow.CFGNode
-import cacophony.controlflow.Register
-import cacophony.controlflow.RegisterLabel
+import cacophony.controlflow.*
 
 class CacophonyPattern {
+    /*****************************
+     * CHAPTER 1. VALUE PATTERNS *
```
mhorod, line 8: Maybe we can split these patterns into separate files? They don't really have to belong to `CacophonyPattern` class.

There's going to be a lot of them, so imo we can start organizing this already.

src/main/kotlin/cacophony/codegen/patterns/CacophonyPattern.kt
```diff
@@ -1,22 +1,215 @@
...
-        val rhsRegisterLabel = RegisterLabel()
+        val lhsLabel = ValueLabel()
+        val rhsLabel = ValueLabel()
+    }
+
+    abstract class UnaryOpPattern {
+        val childLabel = ValueLabel()
+    }
+
+    // TODO: reduce boilerplate (typed DSL?)
```
mhorod, line 20: To give you some inspiration: https://github.com/mhorod/cacophony/blob/a0ffd9363123be86c472f62813b023707d15d1ad/src/test/kotlin/cacophony/TestCfgUtils.kt#L100

I think a great improvement would be to move/copy the node building methods from tests and add the ones that are missing, so we could write something like
```
val tree = registerSlot(lhsLabel) add mem(add(rbp), integer(8))
```

mhorod (commented):
Nice patterns :D

src/main/kotlin/cacophony/codegen/patterns/CacophonyPattern.kt
AHGIJMKLKKZNPJKQR: We can use marker interfaces to group them semantically as well, it might help in the future

mhorod (approved):
LGTM 👍

PR #154, opened 2024-11-22
Add interface for specifying jump condition

honvex-cool:
Now the linearization algorithm will be able to choose whether the jump is to be taken when the condition evaluates to `true` or when it evaluates to `false`


mhorod (approved):
LGTM 👍

PR #155, opened 2024-11-22
Add some patterns and reorganize

mhorod:


src/main/kotlin/cacophony/codegen/patterns/cacophonyPatterns/SideEffectPatterns.kt
```diff
@@ -5,142 +5,145 @@ import cacophony.codegen.patterns.SideEffectPattern
...
-        }
+    override fun makeInstance(fill: SlotFill): List<Instruction> {
+        TODO("Not yet implemented")
     }
+}
 
-    object MultiplicationAssignmentRegisterPattern : SideEffectPattern, RegisterAssignmentTemplate() {
-        override val tree = lhsSlot subeq rhsSlot
+object MultiplicationAssignmentRegisterPattern : SideEffectPattern, RegisterAssignmentTemplate() {
+    override val tree = lhsSlot subeq rhsSlot
```
honvex-cool, line 60: This should be `muleq`, probably just a copy-paste mistake

src/main/kotlin/cacophony/codegen/patterns/cacophonyPatterns/PatternTemplates.kt
```diff
@@ -0,0 +1,31 @@
...
+    val lhsSlot = CFGNode.RegisterSlot(lhsRegisterLabel)
+    val rhsSlot = CFGNode.ValueSlot(rhsLabel)
+}
+
+abstract class MemoryAssignmentTemplate {
+    val lhsLabel = ValueLabel()
+    val rhsLabel = ValueLabel()
+    val lhsSlot = CFGNode.MemoryAccess(CFGNode.ValueSlot(lhsLabel))
+    val rhsSlot = CFGNode.ValueSlot(rhsLabel)
+}
```
honvex-cool, line 19: Probably not a relevant comment for now, but something to keep in mind: when the number of patterns really grows, we could think of writing some code (maybe using reflection?) to generate matching pairs of instances for register and memory access patterns

src/main/kotlin/cacophony/codegen/patterns/cacophonyPatterns/ValuePatterns.kt
```diff
@@ -5,131 +5,136 @@ import cacophony.codegen.patterns.SlotFill
...
+        SubtractionPattern,
+        MultiplicationPattern,
+        DivisionPattern,
+        ModuloPattern,
+        EqualsValuePattern,
+        NotEqualsValuePattern,
+        LessValuePattern,
+        GreaterValuePattern,
+        LessEqualValuePattern,
+        GreaterEqualValuePattern,
```
honvex-cool, line 21: NIT: In `ConditionPatterns.kt`, these are listed in a different order: lt, leq, gt, geq. I think it would be nice to keep those orders aligned. Personally, I like the order from `ConditionPatterns.kt` more, but it's not a strong opinion

src/main/kotlin/cacophony/codegen/patterns/cacophonyPatterns/ValuePatterns.kt
```diff
@@ -5,131 +5,136 @@ import cacophony.codegen.patterns.SlotFill
...
+object GreaterEqualValuePattern : ValuePattern, BinaryOpPattern() {
+    override val tree = lhsSlot geq rhsSlot
 
-        override fun makeInstance(fill: SlotFill, destination: Register): List<Instruction> {
-            TODO("Not yet implemented")
-        }
+    override fun makeInstance(fill: SlotFill, destination: Register): List<Instruction> {
+        TODO("Not yet implemented")
     }
+}
```
honvex-cool, line 124: NIT: As above, I think that it would be nice to use the same order of these operators in both `ConditionPatterns.kt` and here

src/main/kotlin/cacophony/codegen/patterns/cacophonyPatterns/ConditionPatterns.kt
```diff
@@ -1,3 +1,80 @@
...
+    }
+}
+
+object LogicalNotConditionPattern : ConditionPattern, UnaryOpPattern() {
+    override val tree = not(childSlot)
+
+    override fun makeInstance(fill: SlotFill, destinationLabel: BlockLabel, jumpIf: Boolean): List<Instruction> {
+        TODO("Not yet implemented")
+    }
+}
```
honvex-cool, line 80: I'm not saying that it's wrong, just wondering whether we currently need it. Isn't `logical not` encoded in the structure of the CFG? If I recall correctly, when we have a negated condition, we actually evaluate the plain condition and swap the branches. What do you think?

honvex-cool (changes requested):
Excellent reorganization, thanks! :D
I added some comments, but some of them are really only an invitation to a discussion. Overall, I like the commit a lot and it doesn't require many changes :D

honvex-cool (approved):
LGTM :+1:

PR #160, opened 2024-11-24
InstructionMatcher implementation

Giselus:
Implementation still lacks tests.

### DIFF
--- a/src/main/kotlin/cacophony/codegen/instructions/matching/InstructionMatcher.kt
+++ b/src/main/kotlin/cacophony/codegen/instructions/matching/InstructionMatcher.kt
@@ -80,11 +80,11 @@ class InstructionMatcherImpl(
     )
 
     // Internal function to save some boilerplate from three exported functions.
-    // instructionMaker is truly instructionMakerMaker, as it's some helper to generate instructionMakers.
+    // As only thing that is different is instruction maker, it accepts createInstructionMaker to simplify.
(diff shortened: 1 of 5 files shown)


src/main/kotlin/cacophony/codegen/instructions/matching/InstructionMatcher.kt
```diff
@@ -11,3 +12,120 @@ interface InstructionMatcher {
...
+        val toFill: MutableMap<ValueLabel, CFGNode> = mutableMapOf(),
+        var size: Int = 0,
+    )
+
+    // Internal function to save some boilerplate from three exported functions.
+    // instructionMaker is truly instructionMakerMaker, as it's some helper to generate instructionMakers.
+    private fun <PatternType : Pattern> findMatchesWithInstructionMaker(
+        node: CFGNode,
+        patterns: List<PatternType>,
+        instructionMaker: (MatchMetadata, PatternType) -> InstructionMaker,
```
honvex-cool, line 87: I like the `instructionMakerMaker` comment :)
However, in all seriousness, how do you feel about renaming this parameter to `createInstructionMaker` or something similar, with a verb in the name? Maybe this would better indicate the purpose of the argument?

honvex-cool (changes requested):
The implementation looks good so far, I just had one tiny comment. From what I understand, you're still working on some tests, right? If so, please add them and I'll look at the PR again

honvex-cool (approved):
LGTM :+1:

PR #163, opened 2024-11-25
Add prologue and epilogue

FilipKon13:


FilipKon13:
No tests, no pipeline update

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -232,6 +237,118 @@ class FunctionHandlerImpl(
...
+                variables
+                    .map { getVariableAllocation(it) }
+                    .filterIsInstance<VariableAllocation.OnStack>()
+                    .sortedBy { it.offset }
+            run {
+                // Sanity checks
+                var offset = 0
+                for (variable in stackVariables) {
+                    if (variable.offset != offset)
+                        throw CompileException("Holes in stack")
```
KacperTopolski, line 277: CompileException should be used for reporting errors in compiled programs, but this on the first glance looks like an error in a compiler?

FilipKon13: Right, got confused

FilipKon13: Fixed, but found something similar in the `generateFunctionHandlers` function, not sure if this is a similar issue

FilipKon13:
Still fighting with the `CFG of if with variable condition` test and return statements, but it should be the final version more or less

src/test/kotlin/cacophony/controlflow/generation/CFGGenerationTest.kt
```diff
@@ -38,39 +29,32 @@ class CFGGenerationTest {
...
-            CFGFragment(
-                mapOf(
-                    entryLabel to CFGVertex.Jump(writeUnitToRax, returnLabel),
-                    returnLabel to CFGVertex.Final(CFGNode.Return),
-                ),
-                entryLabel,
-            )
-        assertEquivalent(cfg, mapOf(fDef to expectedCFG))
+            cfg {
+                fragment(fDef, listOf(VariableAllocation.OnStack(0)), 8) {
```
honvex-cool, line 34: You are defining `argStack` in this file but not using it - please use it

honvex-cool (changes requested):
Phew, that was quite a challenge to modify all these tests, thanks for the effort! :D
I know that Cacophony's semantics of function function handling is still ~~a mess~~ not completely clear, so all this is something to discuss as a group.
For now I think it will be ok to merge this once the minor comments are resolved.

honvex-cool (approved):
LGTM :+1:

PR #174, opened 2024-11-26
fix noop loop

mhorod:


KacperTopolski:
Add test(s). I think that just adding `while true do true;` to examples is enough, as this is a small fix.

KacperTopolski (approved):

PR #178, opened 2024-11-28
Function call support

Rafi0221:
<s>I've done a little bit of hacking here.

On the one hand `Call` is a `SideEfectPattern`, on the other it needs a `destinationLabel`. Moreover, `CFGNode.Call` is a data class containg a function declaration and we don't want to create a pattern for every function (possibly we can introduce new `LabelSlot`).

To not change interfaces, right now matcher fetches `destinationLabel` from CFG and passes it to `SlotFill`. </s>

New label was introduced.

I also added labels at the beginning of every `BasicBlock`, but changing tests for linearization is not trivial, so these two lines are commented and I am leaving this for @AHGIJMKLKKZNPJKQR 😉.

### DIFF
--- a/src/main/kotlin/cacophony/codegen/BlockLabel.kt
+++ b/src/main/kotlin/cacophony/codegen/BlockLabel.kt
@@ -1,3 +1,15 @@
 package cacophony.codegen
 
+import cacophony.semantic.syntaxtree.Definition
+import kotlin.math.absoluteValue
+
 data class BlockLabel(val name: String)
+
+fun functionBodyLabel(function: Definition.FunctionDeclaration): BlockLabel {
+    return BlockLabel(
+        when (function.identifier) {
+            "<program>" -> "start"
+            else -> "${function.identifier}_${function.arguments.size}_${function.hashCode().absoluteValue}"
+        },
+    )
+}
--- a/src/main/kotlin/cacophony/codegen/BlockLabel.kt
+++ b/src/main/kotlin/cacophony/codegen/BlockLabel.kt
@@ -1,15 +1,23 @@
 package cacophony.codegen
 
 import cacophony.semantic.syntaxtree.Definition
-import kotlin.math.absoluteValue
 
 data class BlockLabel(val name: String)
 
-fun functionBodyLabel(function: Definition.FunctionDeclaration): BlockLabel {
-    return BlockLabel(
-        when (function.identifier) {
-            "<program>" -> "start"
-            else -> "${function.identifier}_${function.arguments.size}_${function.hashCode().absoluteValue}"
-        },
-    )
+object FunctionBodyLabel {
+    private val functionToId = mutableMapOf<Definition.FunctionDeclaration, Int>()
+    private var lastId = 0
+
+    operator fun invoke(function: Definition.FunctionDeclaration): BlockLabel {
+        if (!functionToId.containsKey(function)) {
+            functionToId[function] = lastId++
+        }
+
+        return BlockLabel(
+            when (function.identifier) {
+                "<program>" -> "start"
+                else -> "${function.identifier}_${function.arguments.size}_id${functionToId[function]}"
+            },
+        )
+    }
 }
--- a/src/main/kotlin/cacophony/codegen/BlockLabel.kt
+++ b/src/main/kotlin/cacophony/codegen/BlockLabel.kt
@@ -1,23 +1,15 @@
 package cacophony.codegen
 
 import cacophony.semantic.syntaxtree.Definition
+import kotlin.math.absoluteValue
 
 data class BlockLabel(val name: String)
 
-object FunctionBodyLabel {
-    private val functionToId = mutableMapOf<Definition.FunctionDeclaration, Int>()
-    private var lastId = 0
-
-    operator fun invoke(function: Definition.FunctionDeclaration): BlockLabel {
-        if (!functionToId.containsKey(function)) {
-            functionToId[function] = lastId++
-        }
-
-        return BlockLabel(
-            when (function.identifier) {
-                "<program>" -> "start"
-                else -> "${function.identifier}_${function.arguments.size}_id${functionToId[function]}"
-            },
-        )
-    }
+fun functionBodyLabel(function: Definition.FunctionDeclaration): BlockLabel {
+    return BlockLabel(
+        when (function.identifier) {
+            "<program>" -> "start"
+            else -> "${function.identifier}_${function.arguments.size}_${function.hashCode().absoluteValue}"
+        },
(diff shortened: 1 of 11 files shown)


src/test/kotlin/cacophony/codegen/instructions/matching/InstructionMatcherTest.kt
```diff
@@ -91,4 +94,19 @@ class InstructionMatcherTest {
...
+    @Test
+    fun `matcher sets function call label`() {
+        val instructionMatcher = InstructionMatcherImpl(emptyList(), listOf(CallPattern), emptyList())
+        val function = Definition.FunctionDeclaration(mockk(), "f", mockk(), listOf(), mockk(), mockk())
+        val node = CFGNode.Call(function)
+
+        val match = instructionMatcher.findMatchesForSideEffects(node).elementAt(0)
+        val instructions = match.instructionMaker(emptyMap())
+
+        assertThat(instructions.size == 1)
```
KacperTopolski, line 107: This does not assert anything (change to assertTrue / add .isTrue()). The same happens two lines below.

src/main/kotlin/cacophony/codegen/instructions/matching/InstructionMatcher.kt
```diff
@@ -125,6 +128,13 @@ class InstructionMatcherImpl(
             for ((nestedNode, nestedPattern) in node.children().zip(pattern.children())) {
                 if (!tryMatch(nestedNode, nestedPattern, matchMetadata)) return false
             }
+
+            if (node is CFGNode.Call) {
```
KacperTopolski, line 132: I am personally in favor of creating a new slot type. It should not significantly increase the amount of code, while preserving consistency in the implementation of patterns.

src/main/kotlin/cacophony/codegen/BlockLabel.kt
```diff
@@ -1,3 +1,15 @@
...
+import cacophony.semantic.syntaxtree.Definition
+import kotlin.math.absoluteValue
+
 data class BlockLabel(val name: String)
+
+fun functionBodyLabel(function: Definition.FunctionDeclaration): BlockLabel {
+    return BlockLabel(
+        when (function.identifier) {
+            "<program>" -> "start"
+            else -> "${function.identifier}_${function.arguments.size}_${function.hashCode().absoluteValue}"
```
KacperTopolski, line 12: I do not like the idea of appending hashCode() to asm labels. I think they should be easy to predict based on the function declaration.

KacperTopolski (changes requested):

src/test/kotlin/cacophony/codegen/instructions/matching/InstructionMatcherTest.kt
mhorod: Or use the assertJ like it was designed and write `assertThat(instructions).hasSize(1)` 😉

Rafi0221: Yeah, I had a massive brain fart while writing this test

src/main/kotlin/cacophony/codegen/BlockLabel.kt
Rafi0221: Now suffix of a label is a unique id

src/main/kotlin/cacophony/codegen/BlockLabel.kt
```diff
@@ -1,3 +1,23 @@
 package cacophony.codegen
 
+import cacophony.semantic.syntaxtree.Definition
+
 data class BlockLabel(val name: String)
+
+object FunctionBodyLabel {
```
KacperTopolski, line 7: I do not think we should use singletons to implicitly pass the id of the function. 

Another idea: function handler is already quite a bloated class, so why not make it also responsible for generating asm labels. FunctionHandler.generateFunctionHandlers looks like a good place for that. We have the list of all functions to take the ids, and list of the ancestors, so we can easily make the label depend on the labels of the parents (if we choose to do so in the future, which would be ideal imo, but it is up for discussion when we introduce the FFI).

This would also have the benefit that the Call CFGNode could contain just a string label. As of now it contains a direct link to AST node, which is quite strange in my opinion.

src/main/kotlin/cacophony/codegen/BlockLabel.kt
```diff
@@ -1,3 +1,23 @@
...
+
+    operator fun invoke(function: Definition.FunctionDeclaration): BlockLabel {
+        if (!functionToId.containsKey(function)) {
+            functionToId[function] = lastId++
+        }
+
+        return BlockLabel(
+            when (function.identifier) {
+                "<program>" -> "start"
+                else -> "${function.identifier}_${function.arguments.size}_id${functionToId[function]}"
```
KacperTopolski, line 19: In my opinion the asm label should not change by simply changing the order of the functions. I thought of something like parentlabel_functionname_argcount when I wrote the comment yesterday, but I just noticed that it may lead to some collisions. 

If you have an idea how to make a label depend only on parent function name, function name & argument counts (or some similar values that do not change based on function order / white spaces etc.) please change it. Otherwise we can leave it as is, and maybe discuss this again in the future once we need it.

KacperTopolski (changes requested):

src/main/kotlin/cacophony/codegen/BlockLabel.kt
Rafi0221: Any label that is generated using only local data can lead to collisions. Maybe lets leave this until someone has a better idea.

src/main/kotlin/cacophony/codegen/BlockLabel.kt
KacperTopolski: I do not want to approve mutable singletons storing global state. Maybe just leave go back to the solution with hashes for now.

src/main/kotlin/cacophony/codegen/BlockLabel.kt
KacperTopolski: That is not true in general. e.g. <parent_label>_[function name length]_[function name]_[argument count] works (but is ugly imo).

src/main/kotlin/cacophony/codegen/BlockLabel.kt
Rafi0221: I think some lambdas in C++ are mangled using a hash, so I will go back to that approach.

We might want to implement a proper name mangling if we decide to go with functional direction in the future.

KacperTopolski (approved):

PR #180, opened 2024-11-30
reorganized code

SebastianSpyrzewski:
Minor reorganization of our code, now it should be easier to navigate through files and I think it will be useful when adding new features.
Mainly:

1. Divided cacophony.semantic into subpackages
2. Split this monstrosity of FunctionHandler.kt into several files

The rest of packages looks fine to me, cacophony.codegen is nicely organized and there is not that much happening in the earlier stages.

As next steps of my task I plan to scan through individual files to see if there is something to clean and maybe add some documentation if needed.


src/test/kotlin/cacophony/semantic/analysis/StaticFunctionRelationsTest.kt
```diff
@@ -320,18 +321,18 @@ class StaticFunctionRelationsTest {
...
-        val funG = functionDeclaration("g", funH)
-        val funJ = functionDeclaration("j", block())
-        val varGUse = variableUse("g")
-        val funI = functionDeclaration("i", block(funJ, call(varGUse)))
-        val funFoo = functionDeclaration("foo", block(varA, funG, funI))
-        val varFooUse = variableUse("foo")
-        val funMain = functionDeclaration("main", call(varFooUse))
-
-        val ast = astOf(funFoo, funMain)
+        val varA = cacophony.semantic.variableDeclaration("a", Empty(cacophony.semantic.mockRange()))
```
FilipKon13, line 324: Is this verbosity necessary?

SebastianSpyrzewski: It was done automatically and I did not notice it - i could either import all of these mock classes or move TestUtils to package cacophony.semantic.analysis package, as they are used almost only in here, but I'm not sure if this will be intuitive

KacperTopolski (approved):

PR #181, opened 2024-11-30
implement spill handling

### LINKED ISSUE #168: Spill handling
Handle the case when the register allocation fails and some of the register get spilled.

km1chno:
Resolves #168 

High level idea is the following:

If register allocation fails with all GPRs, we try to allocate them again with all GPRs minus { R8, R9 } <- these are left for spill handling. The allocation should fail again with some spills

Now we adjust `LoweredCFGFragment` for every function:
- call `allocateFrameVariable()` for every spilled virtual register
- iterate over all instructions in basic blocks
- we detect instruction reading spilled register -> we add some instructions called `prologue` of this instruction, these instructions are responsible for loading spills from memory to registers R8 and R9. Then we substitute spilled registers in the function with R8 and R9,
- analogously, if we detect instruction writing to spilled registers -> we substitute registers in the instruction for R8 and R9 and add an `epilogue` - a list of instructions responsible for saving values of R8 and R9 into memory where spills are stored

### DIFF
--- a/src/main/kotlin/cacophony/codegen/instructions/Instruction.kt
+++ b/src/main/kotlin/cacophony/codegen/instructions/Instruction.kt
@@ -25,10 +25,11 @@ data class MemoryAddress(val base: Register, val index: Register?, val scale: In
 
 interface Instruction {
     val registersRead: Set<Register>
-
     val registersWritten: Set<Register>
 
     fun toAsm(hardwareRegisterMapping: HardwareRegisterMapping): String
+
+    fun substituteRegisters(map: Map<Register, Register>): Instruction
 
     fun isNoop(hardwareRegisterMapping: HardwareRegisterMapping): Boolean = false
 }
--- a/src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ArithmeticInstructions.kt
+++ b/src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ArithmeticInstructions.kt
@@ -70,8 +70,8 @@ data class IMulRegReg(
     )
 
 class Cqo : InstructionTemplates.FixedRegistersInstruction() {
-    override val registersRead: Set<Register.FixedRegister> = setOf(Register.FixedRegister(HardwareRegister.RAX))
-    override val registersWritten: Set<Register.FixedRegister> = setOf(Register.FixedRegister(HardwareRegister.RDX))
+    override val registersRead: Set<Register> = setOf(Register.FixedRegister(HardwareRegister.RAX))
+    override val registersWritten: Set<Register> = setOf(Register.FixedRegister(HardwareRegister.RDX))
 
     override fun toAsm(hardwareRegisterMapping: HardwareRegisterMapping) = "cqo"
 }
--- a/src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ArithmeticInstructions.kt
+++ b/src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ArithmeticInstructions.kt
@@ -69,9 +69,9 @@ data class IMulRegReg(
         "imul",
     )
 
-class Cqo : Instruction {
-    override val registersRead: Set<Register> = setOf(Register.FixedRegister(HardwareRegister.RAX))
-    override val registersWritten: Set<Register> = setOf(Register.FixedRegister(HardwareRegister.RDX))
+class Cqo : InstructionTemplates.FixedRegistersInstruction() {
+    override val registersRead: Set<Register.FixedRegister> = setOf(Register.FixedRegister(HardwareRegister.RAX))
+    override val registersWritten: Set<Register.FixedRegister> = setOf(Register.FixedRegister(HardwareRegister.RDX))
 
     override fun toAsm(hardwareRegisterMapping: HardwareRegisterMapping) = "cqo"
 }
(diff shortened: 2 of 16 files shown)


src/main/kotlin/cacophony/codegen/registers/SpillHandling.kt
```diff
@@ -0,0 +1,105 @@
...
+                            "but only have ${spareRegisters.size} spare registers to use",
+                    )
+                }
+
+                val registersSubstitution: Map<Register, FixedRegister> = usedSpills.zip(spareRegisters).toMap()
+
+                val instructionPrologue = readSpills.flatMap { loadSpillIntoReg(it, registersSubstitution[it]!!) }
+                val instructionEpilogue = writtenSpills.flatMap { saveRegIntoSpill(registersSubstitution[it]!!, it) }
+
+                instructionPrologue +
```
KacperTopolski, line 91: Can you fix this formatting or is it forced by a linter?

src/main/kotlin/cacophony/controlflow/FunctionHandler.kt
```diff
@@ -66,7 +66,7 @@ interface FunctionHandler {
 
     fun generateAccessToFramePointer(other: FunctionDeclaration): CFGNode
 
-    fun allocateFrameVariable(): CFGNode.LValue
+    fun allocateFrameVariable(reg: Register.VirtualRegister): CFGNode.LValue
```
KacperTopolski, line 69: This can be a regular Register, it would make the interface clearer.

src/main/kotlin/cacophony/codegen/instructions/Instruction.kt
```diff
@@ -25,10 +25,11 @@ data class MemoryAddress(val base: Register, val index: Register?, val scale: In
 
 interface Instruction {
     val registersRead: Set<Register>
-
     val registersWritten: Set<Register>
 
     fun toAsm(hardwareRegisterMapping: HardwareRegisterMapping): String
+
+    fun substituteVirtualRegisters(map: Map<Register, Register>): Instruction
```
KacperTopolski, line 32: This is only used with VirtualRegisters, but I think that `substituteRegisters` would be a better name given the function parameters.

KacperTopolski (changes requested):

src/main/kotlin/cacophony/codegen/registers/SpillHandling.kt
```diff
@@ -0,0 +1,105 @@
...
+import cacophony.codegen.linearization.BasicBlock
+import cacophony.codegen.linearization.LoweredCFGFragment
+import cacophony.controlflow.CFGNode
+import cacophony.controlflow.FunctionHandler
+import cacophony.controlflow.Register
+import cacophony.controlflow.Register.FixedRegister
+import cacophony.controlflow.Register.VirtualRegister
+import cacophony.utils.CompileException
+
+class SpillHandlingErrorException(reason: String) : CompileException(reason)
```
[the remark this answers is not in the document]
km1chno: true, removed `CompileException` inheritance and changed name to `SpillHandlingException`

src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ArithmeticInstructions.kt
```diff
@@ -85,6 +85,11 @@ data class IDiv(val reg: Register) : Instruction {
         val hardwareReg = hardwareRegisterMapping[reg]
         return "idiv $hardwareReg"
     }
+
+    override fun substituteVirtualRegisters(map: Map<Register, Register>): IDiv =
+        if (map.containsKey(reg)) {
```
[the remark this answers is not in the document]
km1chno: yes it was only for optimization, you are right that it's not worth it, removed it

src/main/kotlin/cacophony/codegen/registers/SpillHandling.kt
km1chno: fixed

km1chno:
Ok some changes:

## Changes in Register Allocation

Changed `RegisterAllocator` to **not throw** if not allowed register appears in interference graph. Instead - throw at the end if any such register was used for anything else than just allocating it to itself. So if `R8` appears in interference, it is fine as long as register allocation says `R8 -> R8` and there is no other `* -> R8`. 

### Why

For spill handling we need `2` spare registers, and we call register allocation with `allowedRegisters = allGPRs \ spareRegisters`. It is fine if spare registers appear in the interference graph, we just want no virtual registers allocated to them. And **they will appear in the interference graph** - `CallInstruction` is using all registers apart from call preserved registers, and call preserved registers are used in instructions in function call prologues and epilogues. So we are physically unable to choose spare register in such a way that they will surely not appear in interference graph! Therefore we need to be ready for that.

## Problem with interference of new instructions with spare registers

After the allocation, we use spare registers to adjust basic blocks to handle spills. We must be careful though, because spare registers must live right before and right after the operation with spills that they are handling. Therefore, every spilled instruction must be handled using spare registers it is not using. 

Here comes the choice of spare registers - I chose `R8` and `R9`, because I believe that they are only used by `CallInstruction`. But `CallInstruction` does not use any virtual registers, so it will never have to be spill handled, so this choice seems optimal.

If in the future this assumption will not be valid anymore, we should obtain an exception that there are not enough spare registers for spill handling. We will then just have to add another one to use 3 spare registers.

KacperTopolski (approved):

PR #183, opened 2024-12-01
Unified the call convention for the static links

FilipKon13:
The static links are now passed as the last (implicit) argument


KacperTopolski (approved):
Please rebase/merge main to check if CI still passes before merging.

PR #185, opened 2024-12-01
Compile and link to executable

AHGIJMKLKKZNPJKQR:


src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -235,4 +237,42 @@ class CacophonyPipeline(
...
+        Files.newBufferedWriter(dest, StandardOpenOption.CREATE).use { writer ->
+            defs.values.forEach {
+                writer.write(it)
+                writer.newLine()
+            }
+        }
+    }
+
+    fun compile(src: Path, dest: Path) {
+        System.out.println("Compiling generated assembly")
```
FilipKon13, line 252: Could we use our logger here instead of System.out?

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -235,4 +237,40 @@ class CacophonyPipeline(
...
+                writer.newLine()
+            }
+        }
+    }
+
+    fun compile(src: Path, dest: Path) {
+        val nasm = ProcessBuilder("nasm", "-f", "elf64", "-o", dest.toString(), src.toString()).inheritIO().start()
+        nasm.waitFor().takeIf { it != 0 }?.let { status ->
+            logger?.logFailedAssembling(status)
+            throw CompileException("Unable to assemble generated code")
```
KacperTopolski, line 255: At this stage, nasm/gcc error nearly certainly indicates a bug inside compiler, i.e. this should not be a CompileException.

src/main/kotlin/cacophony/codegen/instructions/GenerateAsm.kt
```diff
@@ -1,13 +1,22 @@
...
-fun generateAsm(instruction: Instruction, registerAllocation: RegisterAllocation): String {
-    return instruction.toAsm(registerAllocation.successful)
-}
+fun generateAsm(block: BasicBlock, registerAllocation: RegisterAllocation): String =
+    ".${block.label().name}:\n" +
+        block
+            .instructions()
+            .filterNot { it.isNoop(registerAllocation.successful) }
+            .map { it.toAsm(registerAllocation.successful) }
+            .filter(String::isNotEmpty)
```
KacperTopolski, line 14: I think we should have a single way to test if instruction is a noop - either via isNoop, or by testing if the returned string is empty - but not both.

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -235,4 +237,40 @@ class CacophonyPipeline(
...
+        gcc.waitFor().takeIf { it != 0 }?.let { status ->
+            logger?.logFailedLinking(status)
+            throw CompileException("Unable to link compiled code")
+        } ?: logger?.logSuccessfulLinking(dest)
+    }
+
+    fun compile(input: Input, src: Path) {
+        val asmFile = withExtension(src.fileName, ".asm")
+        val objFile = withExtension(src.fileName, ".o")
+        val binFile = withExtension(src.fileName, "")
```
KacperTopolski, line 270: This fails on files with no extensions. Personally I see no reason to strip the extension by default (unless we force the input extension to always be the same), imo it is better to just append new extension to the end of old file name.

src/main/kotlin/cacophony/utils/FileUtils.kt
```diff
@@ -0,0 +1,11 @@
+package cacophony.utils
+
+import java.nio.file.Path
+import java.nio.file.Paths
+
+fun withExtension(file: Path, extension: String) =
+    Paths.get(
+        file.toString().let { name ->
+            name.substring(0, name.lastIndexOf('.')) + extension
```
KacperTopolski, line 9: This kind of logic looks suspicious, surely there is a better way to do that.

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -235,4 +237,40 @@ class CacophonyPipeline(
         asm.forEach { (function, asm) -> println("$function generates asm:\n$asm") }
         return asm
     }
+
+    fun generateAsm(input: Input, dest: Path) {
```
KacperTopolski, line 241: This should not be an overload of regular `generateAsm(input)`, their goals are quite different.

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -235,4 +237,40 @@ class CacophonyPipeline(
...
+        val defs = generateAsm(input)
+        Files.newBufferedWriter(dest, StandardOpenOption.CREATE).use { writer ->
```
KacperTopolski, line 243: This can be just `dest.writeLines(defs.values)`.

KacperTopolski (changes requested):

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
AHGIJMKLKKZNPJKQR: Can you be more specific?

src/main/kotlin/cacophony/codegen/instructions/GenerateAsm.kt
AHGIJMKLKKZNPJKQR: I am not sure what the semantics of an empty instruction are (tbh I'm not sure if it's even possible), I just left this to not introduce possibly breaking changes

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
KacperTopolski: One of them generates the asm, the second one generates asm and writes to some file. Maybe saveAsm would be a better name?

AHGIJMKLKKZNPJKQR: I just separated them entirely as `saveAsm` shouldn't be generating anything while something like `generateAndSaveAsm` is verbose

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -240,4 +242,34 @@ class CacophonyPipeline(
         asm.forEach { (function, asm) -> println("$function generates asm:\n$asm") }
         return asm
     }
+
+    fun saveAsm(asm: Collection<String>, dest: Path) {
+        dest.writeLines(asm, Charsets.UTF_8, StandardOpenOption.CREATE)
```
KacperTopolski, line 247: Is this flag (StandardOpenOption.CREATE) necessary? I think this can be a simple dest.writeLines(asm). Also given the new signature this function should be called writeLines or something like that. At this point maybe just inline it in compile to not think about it?

KacperTopolski (approved):

PR #189, opened 2024-12-02
Fix ret/call registers used

AHGIJMKLKKZNPJKQR:


src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ControlflowInstructions.kt
```diff
@@ -77,24 +77,24 @@ data class Jz(override val label: BlockLabel) : InstructionTemplates.JccInstruct
...
-    override val registersRead: Set<Register> =
-        setOf(Register.FixedRegister(HardwareRegister.RSP)).union(
-            SystemVAMD64CallConvention.preservedRegisters().map { Register.FixedRegister(it) },
-        )
-    override val registersWritten: Set<Register> = setOf(Register.FixedRegister(HardwareRegister.RSP))
+data class Call(val function: Definition.FunctionDeclaration) : Instruction {
+    override val registersRead =
+        setOf(Register.FixedRegister(HardwareRegister.RSP)) union
+            SystemVAMD64CallConvention.preservedRegisters().map(Register::FixedRegister)
+    override val registersWritten: Set<Register> =
```
KacperTopolski, line 84: I think RSP should not be included in registersWritten? Does it even matter?

src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ControlflowInstructions.kt
```diff
@@ -77,24 +77,24 @@ data class Jz(override val label: BlockLabel) : InstructionTemplates.JccInstruct
...
+        HardwareRegister
+            .values()
```
KacperTopolski, line 86: Change to .entries

KacperTopolski (changes requested):

src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ControlflowInstructions.kt
AHGIJMKLKKZNPJKQR: I'm pretty sure it doesn't matter, since `ret` is always the final instruction. I don't see how adding rsp here changes anything

KacperTopolski (approved):

PR #191, opened 2024-12-02
Fix function call being evaluated in condition mode

honvex-cool:
Fixes CFG connectivity when function call is used as a condition in `if` or `while`


mhorod (approved):
lgtm 👍

KacperTopolski (approved):

PR #193, opened 2024-12-02
Fix unconnected `NoOp`

honvex-cool:
When `return` or `break` was visited in `Conditional` mode, the artificial exit was not properly connected
Hopefully, it is fixed now


mhorod (approved):
lgtm 👍

PR #197, opened 2024-12-03
Interfaces for add-ffi

jjaworska:
Interfaces for the "add-ffi" functionality

### DIFF
--- a/src/main/kotlin/cacophony/controlflow/functions/FunctionCallHandlerImpl.kt
+++ b/src/main/kotlin/cacophony/controlflow/functions/FunctionCallHandlerImpl.kt
@@ -1,0 +1,32 @@
+package cacophony.controlflow.functions
+
+import cacophony.controlflow.CFGNode
+import cacophony.controlflow.CFGNode.RegisterUse
+import cacophony.controlflow.HardwareRegister
+import cacophony.controlflow.Register
+import cacophony.semantic.syntaxtree.Definition.FunctionDeclaration
+
+open class FunctionCallHandlerImpl(
+    val function: FunctionDeclaration,
+    // List of parents' handlers ordered from immediate parent.
+    private val ancestorFunctionHandlers: List<FunctionHandler>,
+) : FunctionCallHandler {
+    // Wrapper for generateCall that additionally fills staticLink to parent function.
+    // Since staticLink is not property of node itself, but rather of its children,
+    // if caller is immediate parent, we have to fetch RBP instead.
+    override fun generateCallFrom(
+        callerFunction: FunctionHandler,
+        arguments: List<CFGNode>,
+        result: Register?,
+        respectStackAlignment: Boolean,
+    ): List<CFGNode> {
+        val staticLinkVar =
+            if (ancestorFunctionHandlers.isEmpty() || callerFunction === ancestorFunctionHandlers.first()) {
+                RegisterUse(Register.FixedRegister(HardwareRegister.RBP))
+            } else {
+                callerFunction.generateAccessToFramePointer(ancestorFunctionHandlers.first().getFunctionDeclaration())
+            }
+
+        return generateCall(function, arguments + mutableListOf(staticLinkVar), result, respectStackAlignment)
+    }
+}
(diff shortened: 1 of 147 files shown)


src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -112,14 +112,34 @@ sealed class Definition(
...
         val returnType: Type,
+    ) : Definition(range, identifier) {
+        override fun isEquivalent(other: Expression?): Boolean =
+            super.isEquivalent(other) &&
+                other is FunctionDeclaration &&
+                areEquivalentTypes(type, other.type) &&
+                areEquivalentTypes(returnType, other.returnType)
+    }
+
+    class ForeignFunctionDef(
```
Giselus, line 128: I see that we have a problem with naming and we swapped meanings of declaration and definition. Foreign functions can only be declared, but not defined. We could probably fix it and swap all declarations and definitions, if you don't want to do it here you may keep it as it is for integrity (and maybe order someone else to fix naming 😈).

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -112,14 +112,34 @@ sealed class Definition(
...
+        range: Pair<Location, Location>,
+        identifier: String,
+        type: Type.Functional,
+        returnType: Type,
+    ) : FunctionDef(range, identifier, type, returnType), TreeLeaf
```
Giselus, line 133: We can't have own function declarations in code, but we can have foreignFunction declarations, so maybe we can add toString for printing. I'm unsure if foreign function declarations will be present in AST, so I don't know if my comment makes sense.

src/main/kotlin/cacophony/controlflow/functions/FunctionCallHandlerImpl.kt
```diff
@@ -0,0 +1,32 @@
+package cacophony.controlflow.functions
+
+import cacophony.controlflow.CFGNode
+import cacophony.controlflow.CFGNode.RegisterUse
+import cacophony.controlflow.HardwareRegister
+import cacophony.controlflow.Register
+import cacophony.semantic.syntaxtree.Definition.FunctionDeclaration
+
+open class FunctionCallHandlerImpl(
```
Giselus, line 9: If I understand, we'll have another implementation like _ForeignFunctionCallHandlerImpl_, so maybe _FunctionCallHandlerImpl_ can get a better name that shows that it's user created function? But I'm not sure, just a suggestion.

Giselus (approved):
Looks good overall. I approve, but please think if you want to change something according to comments.

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
jjaworska: that was actually quite bad. I changed the naming, but this makes the diff much bigger

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
jjaworska: Yes, definitely, I added both this and isEquivalent

src/main/kotlin/cacophony/controlflow/functions/FunctionCallHandlerImpl.kt
jjaworska: I was thinking there was one FunctionCallHandlerImpl which just checks inside which function it uses. Maybe let's leave it up to decide for the person who will be implementing this?

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
jjaworska: but of course, it'll change snapshots once again :upside_down_face:

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -125,29 +125,37 @@ sealed class Definition(
                 areEquivalentTypes(returnType, other.returnType)
     }
 
-    class ForeignFunctionDef(
+    class ForeignFunctionDeclaration(
         range: Pair<Location, Location>,
         identifier: String,
         type: Type.Functional,
         returnType: Type,
```
Giselus, line 132: Just a question, I see it's done this way in different places as well. Type.Functional already has returnType, so it seems like we have duplicate returnType here. Maybe one of them is not used, or something. It has too many occurrences in the code to check, so I'll just leave this comment and you may add this as an issue or maybe I don't understand something here.

Giselus (approved):
LGTM 👍

PR #198, opened 2024-12-03
interfaces for structure-prep

jjaworska:


Giselus (approved):
If I understand, Tomek is responsible for majority of interfaces and such.
I guess we also don't need to add anything more, since nobody will try to use structs in the pipeline after typechecker. So it looks weird, because there is nothing there, but it's probably correct.

PR #201, opened 2024-12-03
Add special sideEffect pattern for spills

Giselus:
Special pattern, so that spills handling doesn't create new virtual registers.


km1chno (approved):
goooood, i checked its working trust me

PR #210, opened 2024-12-03
Add structures to AST and grammar

AHGIJMKLKKZNPJKQR:


src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -50,12 +58,21 @@ sealed class Type(
                 areEquivalentTypes(argumentsType, other.argumentsType) &&
                 areEquivalentTypes(returnType, other.returnType)
     }
+
+    class Struct(range: Pair<Location, Location>, val fields: Map<String, Type>) : Type(range) {
```
SebastianSpyrzewski, line 62: maybe this could be `Type.Structural` to match naming convention with `Type.basic` and `Type.Functional`

SebastianSpyrzewski (changes requested):
I see no support for access to struct fields, starting from no lexer token '.' - will it be implemented later?

AHGIJMKLKKZNPJKQR:
@SebastianSpyrzewski 
> I see no support for access to struct fields, starting from no lexer token '.' - will it be implemented later?
Yeah, I decided to separate that to a second PR, to not make this one too bulky and it seems to be a clean way to separate the code

src/main/kotlin/cacophony/semantic/names/NameResolver.kt
```diff
@@ -201,7 +201,11 @@ fun resolveNames(root: AST, diagnostics: Diagnostics): NameResolutionResult {
                 traverseAst(node.rhs, true)
             }
 
-            else -> {}
+            is Struct -> {
+                node.fields.values.forEach { traverseAst(it, true) }
```
jjaworska, line 205: It seems odd to me that we don't use field identifiers here?

jjaworska (commented):
Overall looks great! Really great refactors! but I have this one question below

src/test/kotlin/cacophony/semantic/names/NameResolverTest.kt
```diff
@@ -135,6 +135,28 @@ class NameResolverTest {
                 verify { diagnostics wasNot Called }
             }
 
+            @Test
+            fun `variables in struct`() {
+                // let c = 1;
+                // let a = {x = let b = 2, y = c};
```
jjaworska, line 141: maybe add a bUse?

jjaworska (approved):
Great contribution! :confetti_ball:

PR #211, opened 2024-12-03
fix spill handling

km1chno:
Fixes issue where generated assembly code had instructions with `null` registers after spill handling.

The root cause was that instruction covering was covering nodes with memory access to spill locations using new virtual registers that were not allocated to hardware registers.

The solution is to add patterns to covering that allow doing it without creating new virtual registers, and also adding method `coverWithInstructionsWithoutVirtualRegisters()` to instruction covering to make a promise on interface level that the returned covering is indeed not using any virtual registers.


src/main/kotlin/cacophony/codegen/instructions/CacophonyInstructionCovering.kt
```diff
@@ -22,18 +22,29 @@ class CacophonyInstructionCovering(private val instructionMatcher: InstructionMa
...
                 .union(instructionMatcher.findMatchesForValue(node, Register.VirtualRegister()))
         val bestMatch = matches.maxByOrNull { match -> match.size } ?: error("No match found for $node, ${node.javaClass}")
         return coverGivenMatch(bestMatch)
     }
 
+    override fun coverWithInstructionsWithoutVirtualRegisters(node: CFGNode): List<Instruction> {
+        val matches = instructionMatcher.findMatchesForSideEffects(node)
+        val bestMatch =
+            matches
+                .filter { it.toFill.isEmpty() }
```
Giselus, line 43: I kinda don't like it. toFill is not empty if a pattern has valueSlot, so we check patterns that we know we'll just yank out in this filter. It's not problematic, but maybe you'll come up with better solution.

src/main/kotlin/cacophony/codegen/instructions/CacophonyInstructionCovering.kt
```diff
@@ -22,18 +22,29 @@ class CacophonyInstructionCovering(private val instructionMatcher: InstructionMa
...
+                .maxByOrNull { match -> match.size } ?: error("No match without virtual registers found for $node, ${node.javaClass}")
+        return coverGivenMatch(bestMatch)
+    }
+
     override fun coverWithInstructionsAndJump(node: CFGNode, label: BlockLabel, jumpIf: Boolean): List<Instruction> {
         val matches = instructionMatcher.findMatchesForCondition(node, label, jumpIf)
         val bestMatch = matches.maxByOrNull { match -> match.size } ?: error("No match found for $node, ${node.javaClass}")
```
Giselus, line 50: We can also change this error's message as in methods above.

Giselus (changes requested):

src/main/kotlin/cacophony/codegen/instructions/CacophonyInstructionCovering.kt
km1chno: we could create sub-interface for patterns `NoTemporaryRegistersPattern`, but I'm not sure which patterns can be there?

Or maybe broader question, because I don't know this logic too good - when exactly does a match not contain any temporary virtual registers in it?

Giselus: A pattern can have 3 types of slots: valueSlot, constantSlot and registerSlot. If it has valueSlot it means that it can accept any sort of value and this value may be result of some other computations. And suppose this result is stored in some register for example, we cannot directly use this register, because we may have second children and their computation will change this register, so we have to save it to some virtualRegister. So short answer is that if pattern has any valueSlot, then it will create virtualRegister for result from subtree.

But it could probably be optimized, if result of some subtree is in virtualRegister, we can just use it instead of creating a new one. Maybe after every match that computes something we have virtualRegister and therefore we can **always** reuse it later instead of creating new one. I'm not the author of patterns nor instructionCovering :[ 

You can look into it, or ask some other people.

src/main/kotlin/cacophony/codegen/instructions/CacophonyInstructionCovering.kt
km1chno: done

src/main/kotlin/cacophony/codegen/instructions/CacophonyInstructionCovering.kt
km1chno: thanks for clarification, I created new interface

```
interface NoTemporaryRegistersPattern : SideEffectPattern
```

and dedicated methods in covering and matcher

src/main/kotlin/cacophony/codegen/instructions/matching/InstructionMatcher.kt
```diff
@@ -53,10 +55,26 @@ class InstructionMatcherImpl(
...
-    override fun findMatchesForCondition(node: CFGNode, destinationLabel: BlockLabel, jumpIf: Boolean): Set<Match> {
-        return findMatchesWithInstructionMaker(
+    override fun findMatchesWithoutTemporaryRegisters(node: CFGNode): Set<Match> =
+        findMatchesWithInstructionMaker(
+            node,
+            noTemporaryRegistersPatterns,
+        ) { metadata: MatchMetadata, pattern: SideEffectPattern ->
+            { mapping: ValueSlotMapping ->
+                pattern.makeInstance(
+                    SlotFill(
```
jjaworska, line 66: can we use this opportunity to extract this SlotFill(...) into a separate function? It's quite repetetive now

src/main/kotlin/cacophony/codegen/patterns/cacophonyPatterns/SideEffectPatterns.kt
jjaworska, line 1: I just want to make sure - we do want to say that both patterns like AdditionAssignmentMemoryPattern, which explicitly create a virtual register and patterns like PushPattern, which only might use a virtual register created from a child pattern are "TemporaryRegisterPatterns"?

src/test/kotlin/cacophony/codegen/instructions/matching/InstructionMatcherTest.kt
jjaworska, line 1: is it possible to write a test which demonstrates the bug addressed?

jjaworska (changes requested):

src/main/kotlin/cacophony/codegen/patterns/cacophonyPatterns/SideEffectPatterns.kt
km1chno: yes, my intention was for `NoTemporaryRegistersPattern` to be a interface-level **promise** that for 100% it does not consist of any temporary virtual registers

src/test/kotlin/cacophony/codegen/instructions/matching/InstructionMatcherTest.kt
km1chno: As unit test, I'm not sure it's worth it. This is not really a technical bug, but mismatch in interfaces contracts. The same way there are no tests for covering side effects specifically, I'm not sure how tests for matching with new type of patterns could be easily implemented.

Instead, I added regression test for the following program:
```
let f = [x: Int] -> Int => (
    let a = 1;
    let b = 2;
    let c = 3;
    let d = 4;
    let e = 5;
    let f = 6;
    let g = 7;
    let h = 8;
    let i = 9;
    let j = 10;
    let k = 11;
    let l = 12;
    let m = 13;
    let n = 14;
    let o = 15;
    let p = 16;
    let r = 17;
    let s = 18;
    let t = 19;
    return a+b+c+d+e+f+g+h+i+j+k+l+m+n+o+p+r+s+t;
);
return f[1];
```

that compiles after this PR (and was not compiling before, obviously). It even returns with the correct exit code:D (190)

jjaworska: that looks good!

jjaworska (approved):
Good to go :+1:

src/main/kotlin/cacophony/codegen/instructions/CacophonyInstructionCovering.kt
```diff
@@ -22,21 +22,32 @@ class CacophonyInstructionCovering(private val instructionMatcher: InstructionMa
...
+            matches.maxByOrNull { match -> match.size } ?: error("No match without temporary registers found for $node, ${node.javaClass}")
+        return coverGivenMatch(bestMatch)
+    }
+
     override fun coverWithInstructionsAndJump(node: CFGNode, label: BlockLabel, jumpIf: Boolean): List<Instruction> {
         val matches = instructionMatcher.findMatchesForCondition(node, label, jumpIf)
-        val bestMatch = matches.maxByOrNull { match -> match.size } ?: error("No match found for $node, ${node.javaClass}")
+        val bestMatch =
+            matches.maxByOrNull { match -> match.size }
+                ?: error("No match with instructions for condition found for $node, ${node.javaClass}")
```
Giselus, line 50: A little weird message, every match is "with instructions" as we're covering with instructions. Maybe just "no match for condition" or "no match with jump" or something along these lines.

Giselus (approved):
LGTM 👍

src/main/kotlin/cacophony/codegen/instructions/CacophonyInstructionCovering.kt
km1chno: yeah, changed

PR #222, opened 2024-12-08
Add field access

AHGIJMKLKKZNPJKQR:


src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -4,9 +4,19 @@ import cacophony.utils.Location
...
+sealed interface SyntaxTree {
+    val range: Pair<Location, Location>
 
-sealed interface SyntaxTree
+    fun isEquivalent(other: SyntaxTree?): Boolean = other != null && range == other.range
+}
+
+sealed interface Expression : SyntaxTree, Tree
+
+sealed interface Assignable : Expression
```
jjaworska, line 15: Are there cases in which Assignable is not a TreeLeaf?

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -377,150 +380,151 @@ sealed class OperatorBinary(
...
     }
 
     class Assignment(
         range: Pair<Location, Location>,
-        lhs: Expression,
+        lhs: Assignable,
         rhs: Expression,
-    ) : OperatorBinary(range, lhs, rhs) {
-        override fun isEquivalent(other: Expression?): Boolean = super.isEquivalent(other) && other is Assignment
+    ) : LValueOperator(range, lhs, rhs), Assignable {
```
jjaworska, line 486: should this be Assignable?

jjaworska (changes requested):
Great work!

I'm not sure if changes are required, but I do have a couple of questions.

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
SebastianSpyrzewski: as I unterstand this, FieldRef corresponding to `s.x` is assignable (if s is also assignable) and it has s as its child

AHGIJMKLKKZNPJKQR: Yep, `FieldRef` is not a `TreeLeaf`

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
AHGIJMKLKKZNPJKQR: Imo yes, since `a += b` to me returns `a` after having added `b` to it. It makes stuff like `(a += b) *= c` legal (idk if that's good or bad, but I think that's the way it was before my changes, so I decided not to change the behavior

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
jjaworska: fair

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
jjaworska: ohh I see, it's VariableDeclaration which returns Unit, Assignment returns lhsType. That's right

jjaworska (approved):
Ready to go :+1:

PR #224, opened 2024-12-08
optimize stack allocation for spills

### LINKED ISSUE #207: Debug and stack allocation improvements
- Fix the bug with register spills (occurrence of null registers)
- Implement an optimization of stack allocation:
  1. Adjust the `RegisterAllocation` API to allow for an infinite set of colors
  2. Color the spills registers and use the result to pack stack variables
- If time allows, tackle other bugs or inefficiencies

km1chno:
Resolves #207 

## Changes

- extract first fit graph coloring to generic util
- use it in register allocation (and generally it is refactored a bit)
- reuse it in spill handling to color interference graph for spills with "unlimited number of colors" (equal to number of vertices) to reuse the same frame memory for spills with the same color

### DIFF
--- a/src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
+++ b/src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
@@ -246,7 +246,7 @@ class CacophonyPipeline(
     ): Pair<
         Map<FunctionDeclaration, LoweredCFGFragment>,
         Map<FunctionDeclaration, RegisterAllocation>,
-    > {
+        > {
         if (registerAllocation.values.all { it.spills.isEmpty() }) {
             return covering to registerAllocation
         }
--- a/src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
+++ b/src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
@@ -18,6 +18,7 @@ import cacophony.controlflow.generation.ProgramCFG
 import cacophony.controlflow.generation.generateCFG
 import cacophony.diagnostics.Diagnostics
 import cacophony.grammars.ParseTree
+import cacophony.graphs.FirstFitGraphColoring
 import cacophony.lexer.CacophonyLexer
 import cacophony.parser.CacophonyGrammarSymbol
 import cacophony.parser.CacophonyParser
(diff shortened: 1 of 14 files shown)


src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -268,8 +269,10 @@ class CacophonyPipeline(
                             instructionCovering,
                             functionHandlers[functionDeclaration]!!,
                             loweredCfg,
+                            liveness[functionDeclaration]!!,
```
jjaworska, line 272: probably not a thing to do right now, but a potential refactor in the future - pipeline doesn't seem like the right place for this code to live. What are your thoughts?

src/test/kotlin/cacophony/codegen/registers/SpillHandlingTest.kt
```diff
@@ -476,9 +508,106 @@ class SpillHandlingTest {
...
+                mockLiveness,
                 registerAllocation,
                 setOf(spareReg),
+                mockGraphColoring,
             )
         }
     }
+
+    @Test
+    fun `allocates the same frame memory to spills colored with the same color`() {
```
jjaworska, line 520: This test is already long in terms of lines of code, but describes a quite simple scenario - maybe add e.g. a third register interfering with the other two?

jjaworska (approved):
Great changes! Leave two things for discussion in the comments

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
km1chno: i agree, I hate the current state of pipeline

km1chno: but also I agree that this is a task for future refactor

km1chno: The thing is that spill handling has so many dependencies that it does not fit the "pipeline" paradigm, I'm not sure how this could look better though, perhaps "spill handling" should be a part of internal logic of register allocation

src/test/kotlin/cacophony/codegen/registers/SpillHandlingTest.kt
km1chno: added one reg that interferes with the two, and also one non-spilled reg to check if graph coloring is called with subgraph of interference graph induced by only spilled registers

PR #225, opened 2024-12-08
Added type checking for structures and their fields

SebastianSpyrzewski:
Added support for structures and their fields in typeChecker.


jjaworska (approved):
Well done! I like the refactors

PR #227, opened 2024-12-09
Add io tests

mhorod:


jjaworska (approved):
LGTM :+1:  . You had redundancy with Kacper, but let's stick with your convention

PR #234, opened 2024-12-10
Changed Variable subtypes

SebastianSpyrzewski:


src/main/kotlin/cacophony/controlflow/functions/FunctionHandler.kt
```diff
@@ -16,7 +16,7 @@ interface FunctionHandler {
     fun registerVariableAllocation(variable: Variable, allocation: VariableAllocation)
 
     // Returns static link to parent
-    fun getStaticLink(): Variable.AuxVariable.StaticLinkVariable
+    fun getStaticLink(): Variable
```
FilipKon13, line 19: Maybe: 'TODO: change it to PrimitiveVariable' instead of generic Variable as return type here?

FilipKon13 (approved):

PR #245, opened 2024-12-12
Add creation of variables map

mhorod:


src/main/kotlin/cacophony/semantic/analysis/VariablesMap.kt
```diff
@@ -4,14 +4,199 @@ import cacophony.controlflow.Variable
...
+                /* do nothing */
+            }
+
+            is Definition.FunctionDefinition -> visit(expression.body)
+            is Definition.VariableDeclaration -> visit(expression.value)
+            is Empty -> {
+                /* do nothing */
+            }
+
+            is FieldRef.RValue -> {
```
SebastianSpyrzewski, line 64: I think it fails on something like `(s).x`, as block is not lvalue itself, but can have lvalue inside, which we left unvisited
We should visit expression.obj here as well

src/main/kotlin/cacophony/semantic/analysis/VariablesMap.kt
```diff
@@ -4,14 +4,199 @@ import cacophony.controlflow.Variable
...
+
+    fun visit(expression: Expression) {
+        when (expression) {
+            is FieldRef.LValue -> visitFieldRefLValue(expression)
+
+            is VariableUse -> visitVariableUse(expression)
+
+            is Block -> expression.expressions.forEach { visit(it) }
+            is Definition.FunctionArgument -> {
+                /* do nothing */
```
SebastianSpyrzewski, line 51: a lot of empty cases here, maybe they can be replaced by `is LeafExpression`?

src/test/kotlin/cacophony/semantic/analysis/VariablesMapCreationTest.kt
```diff
@@ -0,0 +1,129 @@
...
+class VariablesMapCreationTest {
+    @Test
+    fun `variable is created for let definitions`() {
+        // given
+        val xDef = variableDeclaration("x", lit(1))
+        val xUse = variableUse("x")
+        val ast = block(xDef, xUse)
+
+        // when
+        val variables = testPipeline().createVariables(ast)
```
SebastianSpyrzewski, line 30: I don't know if usage of `testPipeline()` here is the best, as it calls resolveNames and checkTypes underneath - it shouldn't be much cost to write their result by hand, and we would have isolated test

SebastianSpyrzewski (changes requested):
Nice job, I think it is exactly what was needed to be done, just a few comments here
Also, should we add logging `VariablesMap` to the logger, or do you think it is unnecessary?

src/main/kotlin/cacophony/semantic/analysis/VariablesMap.kt
mhorod: good catch, I totally forgot we have that xD

src/main/kotlin/cacophony/semantic/analysis/VariablesMap.kt
mhorod: good catch, fixed

mhorod:
> Nice job, I think it is exactly what was needed to be done, just a few comments here Also, should we add logging `VariablesMap` to the logger, or do you think it is unnecessary?

Good idea, I added it to the logger :D

SebastianSpyrzewski (approved):
Looks fine, thanks!

PR #246, opened 2024-12-13
liveness 2 (RegistersInteraction)

### LINKED ISSUE #228: Fix Liveness
Currently liveness has some problems

1. It is too restrictive, i.e. sometimes it says there is an interference between registers being copies of one another, but living for a long time and being read at the same time.
2. The copy elimination optimization is causing some problems later, like destroying the stack by assigning to `RSP` https://github.com/mhorod/cacophony/pull/226#discussion_r1876083620

km1chno:
Among other things, resolves #228 

Changes liveness to `RegisterInteraction`, and fixes it


src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ArithmeticInstructions.kt
```diff
@@ -89,6 +89,9 @@ data class IDiv(val reg: Register) : Instruction {
 
     override fun toAsm(hardwareRegisterMapping: HardwareRegisterMapping): String {
         val hardwareReg = hardwareRegisterMapping[reg]
+
+        require(hardwareReg != null) { "No hardware register mapping found for $reg" }
```
SebastianSpyrzewski, line 93: nice idea, I was just thinking if we could do anything to avoid that much code repetition

SebastianSpyrzewski (approved):
Looks nice and clear, good job!
I hope there will be no need for liveness 3

src/main/kotlin/cacophony/codegen/instructions/cacophonyInstructions/ArithmeticInstructions.kt
km1chno: i'm planning to do some refactoring during holidays, also in these parts of code
maybe I'll come up with something clever then

PR #247, opened 2024-12-15
Add code generation for returning structs

MisieqqeisiM:


src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -11,7 +12,7 @@ fun generateCallFrom(
     function: Definition.FunctionDeclaration,
     functionHandler: FunctionHandler?,
     arguments: List<CFGNode>,
-    result: Register?,
+    result: Layout?,
 ): List<CFGNode> =
     when (function) {
         is Definition.ForeignFunctionDeclaration -> {
```
SebastianSpyrzewski, line 18: maybe we could do similar check here for size of return argument?

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -66,10 +72,15 @@ fun generateCall(
     }
 
     nodes.add(CFGNode.Call(function))
-    nodes.add(CFGNode.AdditionAssignment(rsp, CFGNode.ConstantLazy { alignmentShift.value + 8 * stackArguments.size }))
+    nodes.add(CFGNode.AdditionAssignment(rsp, CFGNode.ConstantLazy { alignmentShift.value + stackShift }))
```
SebastianSpyrzewski, line 75: I'm not sure if we want to use stackshift here, I think this addition was to remove all arguments from stack, but we still need access to values returned on stack - it depends on the rest of implementation

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -66,10 +72,15 @@ fun generateCall(
...
+
+    for ((access, register) in registerResults) {
+        nodes.add(CFGNode.Assignment(access as CFGNode.LValue, CFGNode.RegisterUse(Register.FixedRegister(register))))
+    }
 
-    if (result != null)
-        nodes.add(CFGNode.Assignment(CFGNode.RegisterUse(result), CFGNode.RegisterUse(Register.FixedRegister(HardwareRegister.RAX))))
+    for (access in stackResults.reversed()) {
+        nodes.add(CFGNode.Push(access))
```
SebastianSpyrzewski, line 82: why is there a push after the call? do we want pop here instead?

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -66,10 +72,15 @@ fun generateCall(
     }
```
SebastianSpyrzewski, line 73: We pushed necessary arguments to the stack, but [our call convention](https://github.com/mhorod/cacophony/blob/main/DEVELOPMENT.md#function-call-and-return) says that we want to reserve space for results that doesn't fit into single register, and as I unterstand your implementation our EpilogueHandler you use it here and move return values into corresponding places in memory.
I think we need to reserve this space before the call here

src/main/kotlin/cacophony/controlflow/HardwareRegister.kt
```diff
@@ -60,3 +60,14 @@ val REGISTER_ARGUMENT_ORDER =
         HardwareRegister.R8,
         HardwareRegister.R9,
     )
+
+val RETURN_REGISTER_ORDER =
```
SebastianSpyrzewski, line 64: i dont know if we want all of there registers for returned value, but this will be easy to change

MisieqqeisiM: We shall discuss the calling convention further during labs tomorrow. For now this should work.

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -46,7 +37,14 @@ internal class CFGGenerator(
                 }
             }
 
-        val returnVertex = cfg.addFinalVertex(CFGNode.Return)
+        val resultSize =
+            when (function.returnType) {
+                is BaseType.Basic -> 1
+                is BaseType.Structural -> function.returnType.fields.size
```
SebastianSpyrzewski, line 44: that doesn't cover nested structures, we need to calculate size of return type after flattening

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -72,14 +74,23 @@ fun generateCall(
...
 
     nodes.add(CFGNode.Call(function))
-    nodes.add(CFGNode.AdditionAssignment(rsp, CFGNode.ConstantLazy { alignmentShift.value + stackShift }))
+    if (stackResults.isEmpty()) {
```
SebastianSpyrzewski: I think we don't need that much ifology, we can just do `Addition(rsp, 8*stackArguments.size)->read results->Addition(rsp, alignmentShift)` - logic should remain the same

SebastianSpyrzewski (changes requested):
Thank you for fixes, just two more small things

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -66,10 +89,41 @@ fun generateCall(
...
+            CFGNode.AdditionAssignment(
+                rsp,
+                CFGNode.ConstantLazy {
+                    alignmentShift.value + 8 * stackArguments.size +
+                        8 * stackResultsSize
+                },
+            ),
+        )
+    } else {
+        if (stackResultsSize == 0) {
```
SebastianSpyrzewski, line 104: I think we don't need that much ifology, we can just do `Addition(rsp, 8*stackArguments.size)->read results->Addition(rsp, alignmentShift)` - logic should remain the same

FilipKon13:
:oooo:

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
MisieqqeisiM: This would always result in two additions even if they are not necessary, and also break tests :weary:

FilipKon13:
Included in #251

PR #248, opened 2024-12-16
Change VariableDefinition to Variable in AnalyzedFunction

Giselus:


SebastianSpyrzewski (changes requested):
Refactor looks fine, but in terms of logic there are more things yet to be done to support structures in `StaticFunctionRelations` and `VarUseTypeAnalysis`:
- both need to implement visiting `FieldRef` and `Struct` correctly, remembering to visit their subexpressions
- visiting `FieldRef` should mark only the `s.x` Variable as read/written, and not the Variable corresponding to `s`
- on the other hand, reading/writing to structure should mark all of its fields as read/written.
- `visitAssignment` needs changes, as both `StaticFunctionRelations` and `VarUseTypeAnalysis` assume that its lhs is `VariableUse`.

After that we would like to add appropriate tests for structures.

src/main/kotlin/cacophony/semantic/analysis/StaticFunctionRelations.kt
```diff
@@ -119,9 +142,24 @@ private class StaticFunctionsRelationsVisitor {
...
-            else -> visitExpression(expr.lhs)
+        if (expr.lhs is VariableUse || expr.lhs is FieldRef) {
+            val variable = variablesMap.lvalues[expr.lhs as Assignable]!!
+            // Mark all nested variables as written
+            markNestedVariables(variable, VariableUseType.WRITE)
+            // Mark all parents as written
+            var nestedExpression = expr.lhs
+            while (nestedExpression is Assignable) {
+                functionStack.lastOrNull()?.let {
+                    relations[it]?.usedVariables?.add(UsedVariable(variablesMap.lvalues[nestedExpression]!!, VariableUseType.WRITE))
```
SebastianSpyrzewski, line 153: So if we have `s.x.y=1` we mark all the variables `s`, `s.x`, `s.x.y` as used?

src/main/kotlin/cacophony/semantic/analysis/StaticFunctionRelations.kt
```diff
@@ -136,13 +174,13 @@ private class StaticFunctionsRelationsVisitor {
 
     private fun visitVariableWrite(expr: VariableUse) {
```
SebastianSpyrzewski, line 175: I think that is unused now

SebastianSpyrzewski (commented):
FunctionAnalysis looks fine now, good work

src/test/kotlin/cacophony/semantic/analysis/StaticFunctionRelationsTest.kt
```diff
@@ -152,13 +185,130 @@ class StaticFunctionRelationsTest {
...
+        //    a     b
+        //    |    / \
+        //    c   d   e
+        //        |
+        //        f
+        //
+        //
+        // f => ( let s, s.a = s.b.d)
+        val sDeclaration = variableDeclaration("s", Empty(mockRange()))
+        val aDeclaration = variableDeclaration("a", Empty(mockRange()))
```
SebastianSpyrzewski, line 210: our compiler doesn't create any of these declarations besides sDeclaration and we don't need to create them as well, removing them would make code clearer (similarly in VarUseTypeAnalysisTest)

src/test/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysisTest.kt
```diff
@@ -98,21 +112,218 @@ class VarUseTypeAnalysisTest {
...
+            )
+        val funF =
+            unitFunctionDefinition(
+                "f",
+                block(
+                    sDeclaration,
+                    assignment,
+                ),
+            )
+        val ast = astOf(funF)
```
SebastianSpyrzewski, line 188: wrapping in function is not necessary here, we could do just astOf(declaration, assignment)

SebastianSpyrzewski (approved):
I don't see any flaws in program logic, few comments regarding code clarity, but overall I think it's ready to merge

PR #249, opened 2024-12-16
Refactor and add CFG tests

honvex-cool:
Change mocking strategy in CFG tests:
- calls reduced to a single `CFGNode.Call` vertex, if not specified otherwise
- prologue and epilogue replaced with mock `CFGNode`s


SebastianSpyrzewski (approved):
Nice refactor, if all the tests were OK before I see no problems here

PR #250, opened 2024-12-16
Structs IO examples

Rafi0221:


examples/io_not_yet_correct/struct-subtyping-with-functions/program.cac
SebastianSpyrzewski, line 1: I think we could also have example for something like `let f = []->{y : Int} => return {x=2, y=3}`
There will be interesting things going on with function and structures, so we want to make sure to cover them with tests well

SebastianSpyrzewski (changes requested):
Nice and clear tests, take my comment into consideration, besides no cases that we don't cover comes to mind

PR #251, opened 2024-12-17
CFG generation for structs

FilipKon13:
Remaining:
- struct handling in if/else statements (TypeCheckerResult in CFGGenerator is needed for that)
- struct handling in AssignmentOperator (VariablesMap in CFGGenerator is needed for that)
- tests (should work, but who knows, will be added after previous points are resolved)


FilipKon13:
Now, after the VariablesUse map and the type checking result is provided to the CFGGenerator, remaining points can be resolved by adding simple map lookups here and there (all such places are marked with TODO)

FilipKon13:
The new variable allocation (and getVariableAccess method) in the FunctionHandler needs to be merged first

FilipKon13:
Remaining issues:
- epilogue and FunctionHandler still have hardcoded RAX as return register
- generateCall does not handle structs
- no tests

I think all of the above is amongst tasks for the next sprint

FilipKon13:
Kind of ready to push to the main, (#248) is already merged

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -4,47 +4,40 @@ import cacophony.controlflow.*
...
 
         val extended =
             when (bodyCFG) {
                 is SubCFG.Extracted -> extendWithAssignment(bodyCFG, returnValueRegister, EvalMode.Value)
-                is SubCFG.Immediate -> {
-                    val node = CFGNode.Assignment(returnValueRegister, bodyCFG.access)
-                    val vertex = cfg.addUnconditionalVertex(node)
-                    SubCFG.Extracted(vertex, vertex, returnValueRegister)
-                }
+                is SubCFG.Immediate -> assignLayoutWithValue(bodyCFG.access, returnValueRegister, returnValueRegister)
```
SebastianSpyrzewski, line 40: I know we are requiring this to be SimpleLayout for now, but it would be better to change name from returnValueRegister

SebastianSpyrzewski (changes requested):
A lot of good work, nice job!
One little remark and one slightly bigger concern in my review

src/main/kotlin/cacophony/controlflow/functions/FunctionHandlerImpl.kt
```diff
@@ -93,31 +103,23 @@ class FunctionHandlerImpl(
...
-                        function
-                    } else {
-                        ancestorFunctionHandlers.find { it.getStaticLink() == variable }?.getFunctionDeclaration()
-                    }
-                }
-                else -> throw GenerateVariableAccessException(
-                    "Cannot generate access to variables other than static links and source variables.",
-                )
+    override fun generateVariableAccess(variable: Variable.PrimitiveVariable): CFGNode.LValue {
+        if (variable in variableAllocation) { // WARN: hack to handle arguments not used inside function
```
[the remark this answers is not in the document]
FilipKon13: Problem is when we have an Argument which is not used inside the body of the function. Some functionalities was to iterate over all of arguments and do something with that (e.g. prologue generation), and old generateVariableAccess trhows error it this example

`let f = [x:Int] -> Int => 2`

Would throw error in some places.

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
FilipKon13: A lot of stuff here depends on the prologue generation, so there will be more changes.

FilipKon13:
There is actually bigger issue which I am trying to resolve atm, namely the Struct and FieldRef nodes of the AST are not handled in the CFGGenerator. I am working on that and on test.

src/main/kotlin/cacophony/controlflow/functions/FunctionHandlerImpl.kt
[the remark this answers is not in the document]
SebastianSpyrzewski: OK, I see
so I think it should be changed in FunctionAnalysis so that either there is some field `AnalyzedFunction.arguments` or arguments are added to `AnalyzedFunction.variables` even if unused. (we can make our compiler smart and not generate unused function arguments, but I think that would generate more bugs than benefits)
Which version do you prefer? I will make appropriate issue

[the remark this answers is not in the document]
FilipKon13: I meean, it can stay with the additional check that I added, but the preferable solution would be to add them to the `AnalyzedFunction.variables` instead of introducing new place where we have to look for the variables in the function.

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
FilipKon13: Name changed, and removed `require()`s. Still, there are gonna be more changes probably, after the prologue and epilogue generation.

FilipKon13:
Ok, no regression, and even some new tests work. TODOs:
- there is an issue with the variable analysis I think - a subfield accessed in a nested function makes all of the struct to land on the stack (appropriate todo in the relevant test in `StructTest.kt`) - this may be my bad, so I leave opening the issue to you
- no struct arguments and return type tested atm - I would like to merge after (#247) 
- we need to look at the generated assembly sooner or later
- I changed some stuff in the static analysis  (see `Fix visitCompoundAssignment` commit) - someone may want to look at that code, especially given that there were some TODOs left.

Other then that, I think the stuff works more or less.

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
```diff
@@ -145,32 +132,298 @@ class StructTest {
...
+                "bodyEntry" does
+                    jump("bodyExit") {
+                        writeRegister(
+                            getResultRegister(),
+                            memoryAccess(memoryAccess(registerUse(rbp)) sub integer(16)) assign integer(3),
+                        )
+                    }
+            }
+
+        // TODO: expected should be different
```
SebastianSpyrzewski, line 425: I think it is better to make expected CFG is as it should be, and we mark test as Disabled for now

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
```diff
@@ -145,32 +132,298 @@ class StructTest {
...
         assertEquivalent(actualCFG, expectedCFG)
     }
+
+    // TODO: fix this test - accessing subfield makes whole structure land on stack
+    @Test
+    fun `access simple subfield in nested function`() {
+        // given
+
+        /*
+         * let f = [] -> Int => (
```
SebastianSpyrzewski, line 377: I copied this code and try to compile it, got an exception at instruction covering - is this related to the issue you mentioned or is something else yet to be done?

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
```diff
@@ -145,32 +132,298 @@ class StructTest {
...
+                "assign b.b" does jump("reassign b.a") { writeRegister(virC, integer(1)) }
+                "reassign b.a" does jump("reassign b.b") { writeRegister(virB, integer(5)) }
+                "reassign b.b" does jump("assign res") { writeRegister(virC, integer(0)) }
+                "assign res" does jump("bodyExit") { writeRegister(getResultRegister(), registerUse(virB)) }
             }
 
         assertEquivalent(actualCFG, expectedCFG)
     }
+
+    // TODO: fix this test - accessing subfield makes whole structure land on stack
```
SebastianSpyrzewski, line 371: You said that it could be an issue with variable analysis - usage of structure field `s.x` marks both `s` and `s.x` as used in our analysis (but not any other fields of `s`). I am not familiar with CFG generation well enough to know if that is a problem, but maybe you know?

SebastianSpyrzewski (commented):
Few comments regarding the problem you said about, other than that, code looks good

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
FilipKon13: Its more issue of the Variable Allocation in the FunctionHandler I believe

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
FilipKon13: I am not sure how it should look atm, but I will update it before merge (after the other PR's merge as I mentioned)

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
FilipKon13: I have no idea, can you provide more info (DM me, or here)?

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
FilipKon13: Resolved (by you)

src/test/kotlin/cacophony/controlflow/generation/StructTest.kt
FilipKon13: There is a new issue (#253) addressing that.

SebastianSpyrzewski (approved):

FilipKon13:
Includes fixed #247

PR #254, opened 2024-12-28
Add support for combining function calls and field access

SebastianSpyrzewski:
`f[].x` should work now (up to CFG generation at least)

### DIFF
--- a/src/main/kotlin/cacophony/semantic/syntaxtree/ASTGeneration.kt
+++ b/src/main/kotlin/cacophony/semantic/syntaxtree/ASTGeneration.kt
@@ -142,6 +142,32 @@ private fun operatorRegexToAST(children: List<ParseTree<CacophonyGrammarSymbol>>
     }
 }
 
+private fun callAndFieldRegexToAst(children: List<ParseTree<CacophonyGrammarSymbol>>, diagnostics: Diagnostics): Expression {
+    val childNum = children.size
+    if (childNum == 1) {
+        return generateASTInternal(children[0], diagnostics)
+    } else {
+        val newChildren = children.subList(0, childNum - 1)
+        val range = Pair(first = children[0].range.first, second = children[childNum - 1].range.second)
+        val lhs = callAndFieldRegexToAst(newChildren, diagnostics)
+        val lastChild = children[childNum - 1]
+        return if (lastChild is ParseTree.Leaf && lastChild.token.category == VARIABLE_IDENTIFIER) { // field reference
+            if (lhs is Assignable) {
+                FieldRef.LValue(range, lhs, lastChild.token.context)
+            } else {
+                FieldRef.RValue(range, lhs, lastChild.token.context)
+            }
+        } else if (lastChild is ParseTree.Branch && lastChild.production.lhs == FUNCTION_CALL) {
+            val arguments = mutableListOf<Expression>()
+            assert(lastChild.production.lhs == FUNCTION_CALL)
+            for (child in lastChild.children) {
+                arguments.add(generateASTInternal(child, diagnostics))
+            }
+            FunctionCall(range, lhs, arguments)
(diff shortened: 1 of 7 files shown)


src/main/kotlin/cacophony/semantic/syntaxtree/ASTGeneration.kt
```diff
@@ -142,6 +142,32 @@ private fun operatorRegexToAST(children: List<ParseTree<CacophonyGrammarSymbol>>
...
+                FieldRef.LValue(range, lhs, lastChild.token.context)
+            } else {
+                FieldRef.RValue(range, lhs, lastChild.token.context)
+            }
+        } else if (lastChild is ParseTree.Branch && lastChild.production.lhs == FUNCTION_CALL) {
+            val arguments = mutableListOf<Expression>()
+            assert(lastChild.production.lhs == FUNCTION_CALL)
+            for (child in lastChild.children) {
+                arguments.add(generateASTInternal(child, diagnostics))
+            }
```
AHGIJMKLKKZNPJKQR, line 165: **suggestion:** this can be simplified
```suggestion
            assert(lastChild.production.lhs == FUNCTION_CALL)
            val arguments = lastChild.children.map { generateASTInternal(it, diagnostics) }
```

src/main/kotlin/cacophony/semantic/syntaxtree/ASTGeneration.kt
```diff
@@ -142,6 +142,32 @@ private fun operatorRegexToAST(children: List<ParseTree<CacophonyGrammarSymbol>>
     }
 }
 
+private fun callAndFieldRegexToAst(children: List<ParseTree<CacophonyGrammarSymbol>>, diagnostics: Diagnostics): Expression {
+    val childNum = children.size
+    if (childNum == 1) {
+        return generateASTInternal(children[0], diagnostics)
+    } else {
```
AHGIJMKLKKZNPJKQR, line 149: **nitpick:** this `else` is redundant since the `if` has a `return` in it and only makes the code more indented

AHGIJMKLKKZNPJKQR (changes requested):
Mostly small comments, main issue I have is that there is no test for this new behavior in `ASTGenerationTests` (also maybe a snapshot test?)

SebastianSpyrzewski:
I will add tests in `ASTGenerationTests` in a moment.
I'm not sure about snapshots - there is example of behavior `f[].a` in not_yet_correct_directory, so there are no snapshots for it yet - when support for structs will be finished, appropriate snapshots will be added. (they can be added right now, as our snapshots cover only stages up to type checking, but current convention is that we don't have them for not_yet_correct examples).
As for `s.f[]` and `f[][]`, they are not supported for now, so I also don't think we should add examples.

src/test/kotlin/cacophony/semantic/syntaxtree/ASTGenerationTests.kt
```diff
@@ -114,7 +114,7 @@ class ASTGenerationTests {
     @Test
     fun `lexer fail causes ast to not generate`() {
         val diagnostics = computeFailDiagnostics("?1")
-        assertThat(diagnostics).isNotEmpty()
+        assertThat(diagnostics).isNotEmpty
```
AHGIJMKLKKZNPJKQR, line 117: **question:** This looks suspicious, doesn't it remove the assertion?

AHGIJMKLKKZNPJKQR (approved):
one question only so approving to prevent blocking

src/test/kotlin/cacophony/semantic/syntaxtree/ASTGenerationTests.kt
SebastianSpyrzewski: the previous version was getting a warning, and it looks like assertion behaves the same

PR #256, opened 2025-01-02
Syntax for references

SebastianSpyrzewski:
Changes:
-parsing of reference types `&`, allocations `$` and deallocations `@`,
-few simple examples in `not_yet_correct/references`,
-tests for new behavior in `ASTGenerationTest`,
-new kind of tests in the above file checking if type was parsed correctly

### DIFF
--- a/examples/not_yet_correct/references/dereference.cac
+++ b/examples/not_yet_correct/references/dereference.cac
@@ -1,0 +1,5 @@
+let int_reference = $2;
+
+let x: Int = @int_reference;
+
+@int_reference = 5;
--- a/src/test/kotlin/cacophony/semantic/syntaxtree/ASTGenerationTests.kt
+++ b/src/test/kotlin/cacophony/semantic/syntaxtree/ASTGenerationTests.kt
@@ -91,6 +91,16 @@ class ASTGenerationTests {
         return ast ?: throw exc!!
     }
 
+    private fun getInnerBlock(ast: AST): AST { // unwrap AST
+        return ast.children().first().children().first() as Block
+    }
+
+    private fun computeType(type: String): Type? {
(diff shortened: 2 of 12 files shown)


examples/not_yet_correct/references/dereference.cac
AHGIJMKLKKZNPJKQR, line 1: **suggestion:** add a example for structs with reference fields please

src/test/kotlin/cacophony/semantic/syntaxtree/ASTGenerationTests.kt
```diff
@@ -716,4 +726,89 @@ class ASTGenerationTests {
...
+    fun `simple allocation`() {
+        val actual = computeAST("$2")
+        val expected =
+            Allocation(
+                anyLocation(),
+                literal(2),
+            )
+        assertEquivalentAST(mockWrapInFunction(expected), actual)
+    }
+
```
AHGIJMKLKKZNPJKQR, line 740: **suggestion:** add tests for mixing allocations and structs please

AHGIJMKLKKZNPJKQR (approved):
approving preemptively to not block this, but please add those tests

SebastianSpyrzewski:
I decided to change priority of `@` - our idea of "passing assignability through blocks" is not that easy to carry out and now we have easy way to access fields of structure allocated on heap.

Only minor disadvantage is that `@` and `$` have different priorities now, which can be a bit counterintuitive, but tbh makes sense to me - `@x.a` is `(@x).a`, while `$x.a` is `$(x.a)`, but it feels more natural than `($x).a`, which does not compile and in actual usage I think is sensible.
(The same case would be for function calls `@f[]` and `$f[]`).

I will add tests in a moment.

AHGIJMKLKKZNPJKQR (commented):
new changes look ok

PR #266, opened 2025-01-03
CFG support for references

FilipKon13:
This is a hacky initial draft, to see if stuff works and to allow an early feedback


FilipKon13:
Dereference is more or less done. I also fixed some stuff related to using structs in the conditions and added appropriate tests

FilipKon13:
Moved some tests from not_yet_correct to io_test and correct, looks sensible

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -27,7 +27,7 @@ fun generateCallFrom(
         }
 
         is Definition.FunctionDefinition -> {
-            if (function.arguments.size != arguments.size) {
+            if (function.arguments.sumOf { it.type.size() } != arguments.size) {
```
AHGIJMKLKKZNPJKQR, line 30: **question:** Why change this check? Also aren't these conditions enforced by the type checker?

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -77,14 +84,19 @@ internal class CFGGenerator(
...
         when (source) {
             is SimpleLayout -> {
-                require(destination is SimpleLayout) // by type checking
-                require(destination.access is CFGNode.LValue)
-                val write = CFGNode.Assignment(destination.access, source.access)
-                listOf(cfg.addUnconditionalVertex(write))
+                // TODO: restore my sanity
+                if (destination is StructLayout && source.access is CFGNode.NoOp) {
+                    listOf(cfg.addUnconditionalVertex(CFGNode.NoOp))
+                } else {
```
AHGIJMKLKKZNPJKQR, line 90: **question:** what's going on here? What is an assignment to a `SimpleLayout` from a `StructLayout` that can be accessed with no operations :confused: ?

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -134,28 +146,96 @@ internal class CFGGenerator(
...
+            is Dereference -> visitDereference(expression, mode, context)
+            is Assignable -> visitAssignable(expression, mode)
             else -> error("Unexpected expression for CFG generation: $expression")
         }
 
+    private fun visitAllocation(expression: Allocation, mode: EvalMode, context: Context): SubCFG {
+        val calcExpression = visit(expression.value, mode, context)
+        return when (mode) {
+            is EvalMode.Value -> {
+                val type = typeCheckingResult.expressionTypes[expression]!!
```
AHGIJMKLKKZNPJKQR, line 161: **question:** I'm not a 100% on this, but with how the type checker works, but for
```
let x = $(let y: {a: Int} = {a = 0, b = 1}; {c = y})
```
The resulting type of `x` will be `{c: {a: Int}}` and the field `b` will not be allocated on the heap, but will exist temporarily on the stack?

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -134,28 +146,96 @@ internal class CFGGenerator(
...
     }
 
-    private fun wrapExtracted(subCFG: SubCFG): SubCFG.Extracted =
-        when (subCFG) {
-            is SubCFG.Extracted -> subCFG
-            is SubCFG.Immediate -> {
-                val vertex = cfg.addUnconditionalVertex(CFGNode.NoOp)
-                SubCFG.Extracted(vertex, vertex, subCFG.access)
-            }
+    private fun extendWithConditional(fieldAccess: SubCFG, mode: EvalMode.Conditional): SubCFG.Extracted {
```
AHGIJMKLKKZNPJKQR, line 227: **nitpick:** Rename the `fieldAccess` parameter

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -227,29 +320,22 @@ internal class CFGGenerator(
...
-        return if (mode is EvalMode.Conditional) {
-            // by type checking
-            require(resultLayout is SimpleLayout)
-            val conditionVertex = cfg.addConditionalVertex(resultLayout.access neq integer(0))
-            callSequence.exit.connect(conditionVertex.label)
-            conditionVertex.connectTrue(mode.trueEntry.label)
-            conditionVertex.connectFalse(mode.falseEntry.label)
-            SubCFG.Extracted(conditionVertex, mode.exit, CFGNode.NoOp)
-        } else SubCFG.Extracted(entry, callSequence.exit, resultLayout ?: SimpleLayout(CFGNode.NoOp))
+        return if (mode is EvalMode.Conditional) extendWithConditional(callSequence, mode)
```
AHGIJMKLKKZNPJKQR, line 337: **question:** Isn't the layout of `callSequence` a `StructLayout` for functions that return structs? The called function assumes it's a `SimpleLayout`

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
FilipKon13: `arguments` are primitive fields of the arguments, not the number of arguments, this check was incorrect for the struct types. Previously arguments was a list of Layouts, idk why it was changed, I just adjusted it.

Regarding the second question - i am pretty sure they are, this is some internal sanity check i believe, this is not my code

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
FilipKon13: Will exist in a virtual register or as an Immediate. In many situations, it will be removed łater after CFGgeneration. In this particular case, y exists in virtual register, {a = 0, b = 1} exists as Layout inside an Immediate and returned Value (of type c: a: Int) exists in a virtual register from which it is later moved to the memory allocated on the Heap.

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
AHGIJMKLKKZNPJKQR: Let's leave it then

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
FilipKon13: This is not my code actually - this hack is introduced by #270. This tries to fix problem with Assignment from expressions with Void type. Previously, everything (including Void) was an single value, however, now when we try to write something like let x : {a:Int} = return (); we have problem, because return is typed to Void and visiting this statement returns Simple Layout Access (with nop). This no longer works as you can see, we cannot assing to x from such Layout. 

That solves this problem. My proposition is to create VoidLayout which allows to do in the more consistent manner, however this looks like quite a refactor, possibly even a separate task to introduce it and test thoroughly

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
FilipKon13: This is not my code actually - this hack is introduced by #270. This tries to fix problem with Assignment from expressions with Void type. Previously, everything (including Void) was an single value, however, now when we try to write something like let x : {a:Int} = return (); we have problem, because return is typed to Void and visiting this statement returns Simple Layout Access (with nop). This no longer works as you can see, we cannot assing to x from such Layout.

That solves this problem. My proposition is to create VoidLayout which allows to do in the more consistent manner, however this looks like quite a refactor, possibly even a separate task to introduce it and test thoroughly

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
FilipKon13: Removed wrong comment, but i guess you have seen it already

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
AHGIJMKLKKZNPJKQR: okay, that clears this up. If it works then I guess we keep it since adding a new layout type is a bit too much work for right now

AHGIJMKLKKZNPJKQR: This got answered by @FilipKon13

AHGIJMKLKKZNPJKQR (approved):

PR #267, opened 2025-01-03
Add conversion of CFG to builder string

mhorod:


AHGIJMKLKKZNPJKQR (approved):
LGTM, though maybe we could make it possible to generate cfg builders using a Gradle task like we do with graphviz

src/main/kotlin/cacophony/controlflow/print/CFGToBuilder.kt
```diff
@@ -0,0 +1,103 @@
...
+            val register =
+                when (tree.register) {
+                    is Register.FixedRegister -> tree.register.hardwareRegister.toString().lowercase()
+                    is Register.VirtualRegister -> "virtualRegister(\"${tree.register}\")"
+                }
+            "registerUse($register)"
+        }
+        is CFGNode.Comment -> "CFGNode.Comment(\"${tree.comment}\")"
+        is CFGNode.ConstantKnown -> "CFGNode.ConstantKnown(${tree.value})"
+        is CFGNode.ConstantLazy -> "constantLazy"
```
KacperTopolski, line 75: Since `FragmentEquivalenceVisitor` tests constants for value equality, both versions of constant can be converted to `CFGNode.ConstantKnown(${tree.value})`

PR #268, opened 2025-01-04
Create object outlines

Rafi0221:
Additionally, fixed RegisterByte printing in asm.

### DIFF
--- a/src/main/kotlin/cacophony/semantic/rtti/ObjectOutline.kt
+++ b/src/main/kotlin/cacophony/semantic/rtti/ObjectOutline.kt
@@ -1,8 +1,65 @@
 package cacophony.semantic.rtti
 
-import cacophony.semantic.syntaxtree.Type
-import cacophony.semantic.types.TypeExpr
+import cacophony.semantic.types.*
+import kotlin.math.absoluteValue
 
 typealias ObjectOutlineLocation = Map<TypeExpr, String>
 
-fun sizeOf(type: Type): Int = throw NotImplementedError()
+/**
+ * Object outline consists of 8B blocks:
+ * - The first block denotes object size (in 8B blocks, e.g. Int or a reference has size 1).
+ * - The next ceil(size / 64) blocks contain info if a field in the struct is a pointer.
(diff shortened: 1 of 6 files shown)


src/main/kotlin/cacophony/semantic/rtti/ObjectOutline.kt
```diff
@@ -1,8 +1,73 @@
...
+    private fun structToLabel(type: TypeExpr): String =
+        when (type) {
+            is StructType -> "Struct_${type.fields.keys.sorted().joinToString("_")}_${type.hashCode().absoluteValue}"
+            is BuiltinType.BooleanType -> "Boolean_${type.hashCode().absoluteValue}"
+            is BuiltinType.IntegerType -> "Int_${type.hashCode().absoluteValue}"
+            is BuiltinType.UnitType -> "Unit_${type.hashCode().absoluteValue}"
+            is FunctionType -> "Function_${type.hashCode().absoluteValue}"
+            is ReferentialType -> "Ref_${structToLabel(type.type)}"
+            is TypeExpr.VoidType -> "Void_${type.hashCode().absoluteValue}"
+        }
```
AHGIJMKLKKZNPJKQR, line 29: **suggestion:** since nasm labels have a limit on length (4096 characters), I would maybe make these a bit shorter, i.e. instead of `Type_$hashcode` just the first letter of the type without the `_`.

More importantly, I wouldn't use the field names in the label for `StructType` (are these even necessary at all?). If they are necessary, I would instead use the hashcodes of the names to prevent someone getting errors after using super verbose fields like `{supercalifragilisticexpialidocious = 2}`

AHGIJMKLKKZNPJKQR (approved):
looks clean, only nitpicks

src/main/kotlin/cacophony/semantic/rtti/ObjectOutline.kt
Rafi0221: I was not aware of the length limit. The struct fields were added to make debugging a bit easier. Also. I adjusted the labels of deep references, so they have shorter labels.

PR #269, opened 2025-01-04
Add heap read-write analysis

honvex-cool:
Create a special `Variable` representing the heap.

Allocation (`$10`) always writes to `Heap`.
The following examples summarize the analysis of dereferencing:
- `@x + 2` only reads from `Heap`
- `@x = 2` only writes to `Heap`
- `@x += 2` both reads from and writes to `Heap`
All foreign function calls read from and write to `Heap`.

It is worth noting that, in the expression `@x = 2`, also the subexpression `@x` is considered to *write* to `Heap`. I believe that this is consistent with how the subexpression `a` in `a = 5` is considered to write to a variable, even though the actual assignment does not occur in this subexpression. The case with `+=` and read/write access is similar.


src/test/kotlin/cacophony/semantic/syntaxtree/ASTGenerationTests.kt
```diff
@@ -17,7 +17,9 @@ class ASTGenerationTests {
             return
         }
         val expectedStr = TreePrinter().printTree(expected)
+        println(expectedStr)
```
FilipKon13, line 20: Remove println

src/main/kotlin/cacophony/pipeline/CacophonyLogger.kt
```diff
@@ -228,6 +228,8 @@ class CacophonyLogger : Logger<Int, TokenCategorySpecific, CacophonyGrammarSymbo
             is Variable.StructVariable ->
                 "$variable".split("$").last() + " { " +
                     variable.fields.entries.joinToString(", ") { (k, v) -> "$k -> ${variableToString(v)}" } + " }"
+
+            Variable.Heap -> variable.toString()
```
FilipKon13, line 232: Use `is`, similarly in the other places

src/main/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysis.kt
```diff
@@ -175,71 +181,47 @@ private class VarUseVisitor(
...
-            TODO("unimplemented branch for different assignment type")
-//            visitExpression(expr.lhs)
+    private fun visitAssignment(expr: OperatorBinary.LValueOperator, compound: Boolean) {
+        val lhsUseType =
+            if (compound) {
+                VariableUseType.READ_WRITE
+            } else {
+                VariableUseType.WRITE
+            }
+        check(expr.lhs is Assignable) {
```
FilipKon13, line 205: require with IllegalArgumentException seems more appropriate here, but it is only a personal preferencje. BTW, is this condition satisfied by previous stages of compilation, or compiler does not recognize some static reasoning? Either way, comment why this should be satisfied.

FilipKon13 (commented):
Only minor stuff, I leave checking if the task is completed to the przedownik

src/test/kotlin/cacophony/semantic/syntaxtree/ASTGenerationTests.kt
honvex-cool: Done :D

src/main/kotlin/cacophony/pipeline/CacophonyLogger.kt
honvex-cool: Done :D

src/main/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysis.kt
honvex-cool: I changed it to `require` and provided a comment - this is certainly an issue of the compiler not performing in-depth static analysis. I also experimented with opening `val lhs` in `OperatorBinary` and overriding it in `LValueOperator` - this makes the verification redundant but it also changes some snapshots, so imo not worth it atm.

FilipKon13: Yeah, being more specific in the hierarchy seems like a clean solution, but i agree with your last comment

src/main/kotlin/cacophony/controlflow/Variable.kt
```diff
@@ -16,10 +16,13 @@ sealed class Variable(
         constructor(fields: Map<String, Variable>) : this(fields, "sv${index++}")
     }
 
+    object Heap : Variable("<heap>")
+
     fun getPrimitives(): List<PrimitiveVariable> =
         when (this) {
             is PrimitiveVariable -> listOf(this)
             is StructVariable -> fields.map { (_, field) -> field.getPrimitives() }.flatten()
+            Heap -> emptyList()
```
AHGIJMKLKKZNPJKQR, line 25: **suggestion:** elsewhere @FilipKon13 left a comment to use `is` for consistency, so I'd change this too

src/main/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysis.kt
```diff
@@ -315,4 +304,34 @@ private class VarUseVisitor(
...
+    private fun getUnderlyingDereferenceFromChainOfFieldRef(expr: Assignable): Dereference? {
+        var current = expr
+        while (true) {
+            when (current) {
+                is VariableUse -> return null // reached
+                is Dereference -> return current
+                is FieldRef.LValue -> current = current.obj
+            }
+        }
+    }
```
AHGIJMKLKKZNPJKQR, line 336: **question:** Just a quick sanity check: for structs allocated on the heap we don't create variables for each field, right? Asking this here because if that's not the case then this has to be changed I think

src/test/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysisTest.kt
```diff
@@ -605,4 +609,345 @@ class VarUseTypeAnalysisTest {
...
+        assertThat(result).containsAllEntriesOf(
+            mapOf(
+                expr1 to mapOf(pVariable to VariableUseType.READ, Variable.Heap to VariableUseType.READ),
+                expr2 to mapOf(pVariable to VariableUseType.READ, Variable.Heap to VariableUseType.WRITE),
+                expr3 to mapOf(pVariable to VariableUseType.READ, Variable.Heap to VariableUseType.READ_WRITE),
+                expr4 to mapOf(pVariable to VariableUseType.READ, Variable.Heap to VariableUseType.READ),
+            ),
+        )
+    }
+
```
AHGIJMKLKKZNPJKQR, line 927: **praise:** some nice tests:) I think we could add 1 more that checks nesting like `let y = $1; let x = $(@y + 1)`

src/main/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysis.kt
honvex-cool: As discussed with other Collaborators, we do not create variables for stuff on the heap, so I think this should be OK

AHGIJMKLKKZNPJKQR (approved):
approving preemptively since basically no issues

src/main/kotlin/cacophony/controlflow/Variable.kt
honvex-cool: Thanks for catching that, I just missed it. Done :D

src/test/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysisTest.kt
honvex-cool: Thanks, I added the test that you suggested :D

PR #271, opened 2025-01-05
Add type checking for references

### LINKED ISSUE #259: Add type checking for references
Type checking should make sure that

- The result of an allocation has a referential type
- Only expressions that have a referential type are dereferenced
- There is no relation between referential and non-referential types

km1chno:
Resolves #259 

All typechecker changes were already added in #266:D

What I did:
- added subtype check for references (covariance)
- added tests
- voila!


src/main/kotlin/cacophony/semantic/types/TypeChecker.kt
```diff
@@ -268,8 +268,19 @@ private class Typer(
                     BuiltinType.UnitType
                 }
 
-                is Allocation -> throw NotImplementedError()
-                is Dereference -> throw NotImplementedError()
+                is Allocation -> {
+                    val valueType = typeExpression(expression.value) ?: return null
```
AHGIJMKLKKZNPJKQR, line 272: **suggestion:** Maybe this is a good place to prevent allocations of `UnitType`? I don't think `let x = $()` makes sense

SebastianSpyrzewski:
Do we support subtyping for references like that? This way it would be legal to write code like
```
let big_struct_ref = ${x=1, y=2};
let small_struct_ref : &{x: Int} = big_struct_ref;
```
and maybe this is OK, but I'm not sure how garbage collector will deal with that

src/main/kotlin/cacophony/semantic/types/TypeChecker.kt
km1chno: void too?;)

AHGIJMKLKKZNPJKQR: Yeah I think so too, though I think that's impossible to achieve

AHGIJMKLKKZNPJKQR: Actually nvm, you could do `$(return x)`

AHGIJMKLKKZNPJKQR:
I'm pretty sure subtyping is fine since dereferencing from the cfg onwards only cares about rtti

FilipKon13:
I do not think that subtyping on references is fine like that - we generate access to fields based on type of the reference alone, but offsets are different in subtype

km1chno:
@FilipKon13 yeah, so we would have to use RTTI to read struct fields through references - i'll remove the support for subtyping on references, we can do this as another feature

src/main/kotlin/cacophony/semantic/types/TypeChecker.kt
km1chno: done

AHGIJMKLKKZNPJKQR (approved):

PR #272, opened 2025-01-05
Enhance pipeline

### LINKED ISSUE #217: Make pipeline runs configurable
In current form the pipeline is a bit difficult to debug due to the amount of information logged.
It would be nice to be able to configure (e.g. with flags) which pipeline steps should be ran and what should be logged.

Some additional nice to haves:
- feature flags for optimizations/parameters (e.g which patterns and registers to use, or whether to optimize nops in CFG)
- human friendly and computer friendly formats of logs (e.g. we can plot CFG using graphviz, maybe other parts of pipelines could also be translated to slightly more readable format)
- logging to files (possibly separate file for each pipeline step)

jjaworska:
Resolves #217


src/test/kotlin/cacophony/examples/ExamplesTest.kt
```diff
@@ -37,10 +37,14 @@ class ExamplesTest {
...
 
         try {
-            result.ast = pipeline.generateAST(input)
-            result.resolvedVariables = pipeline.resolveOverloads(result.ast!!)
+            result.ast = pipeline.generateAst(input)
+            val nameResolutionResult = pipeline.resolveNames(result.ast!!)
+            result.resolvedVariables = pipeline.resolveOverloads(result.ast!!, nameResolutionResult)
             result.types = pipeline.checkTypes(result.ast!!, result.resolvedVariables!!)
-            result.analysisResult = pipeline.analyzeFunctions(result.ast!!, result.resolvedVariables!!)
+            // result.analysisResult = pipeline.analyzeFunctions(result.ast!!, result.resolvedVariables!!)
```
AHGIJMKLKKZNPJKQR, line 44: **nitpick:** remove commented code

AHGIJMKLKKZNPJKQR (commented):
Very nice changes, :muscle: . 

One issue I found is that `compile.sh` should use `$*` instead of `$1` to work correctly with multiple arguments

Also, it would be nice to have a `-c/--clean` flag though, because if you compile once with `-v` and then only want specific logs it's a bit of a hastle

AHGIJMKLKKZNPJKQR (approved):
approving preemptively, since I left only minor comments

PR #273, opened 2025-01-06
Find references in virtual registers

Giselus:


src/main/kotlin/cacophony/codegen/registers/SpillHandling.kt
```diff
@@ -161,6 +161,17 @@ private fun colorSpills(
...
+            spillsColoring.forEach { (reg, color) ->
+                getOrPut(color) { mutableSetOf() }.add(reg)
+            }
+        }.any {
+            it.value.any { reg -> reg.holdsReference } &&
+                it.value.any { reg -> !reg.holdsReference }
+        }
+    ) {
+        throw SpillHandlingException("Same color assigned to register with and without reference")
+    }
```
AHGIJMKLKKZNPJKQR, line 174: **question:** How is this enforced during coloring? I don't see anything that prevents this at that stage

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -23,7 +24,7 @@ fun generateCallFrom(
             if (result != null && !layoutMatchesType(result, function.type.returnType)) {
                 throw IllegalArgumentException("Wrong result layout")
             }
-            generateCall(function, arguments, result, callerFunction.getStackSpace())
+            generateCall(function, function.type.argumentsType, arguments, result, callerFunction.getStackSpace())
```
AHGIJMKLKKZNPJKQR, line 27: **issue:** I think you want to use the type checking result here instead of taking the type from the AST, which can be null

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -34,20 +35,30 @@ fun generateCallFrom(
...
     function: Definition.FunctionDeclaration,
+    argTypes: List<Type>,
     arguments: List<CFGNode>,
     result: Layout?,
     callerFunctionStackSize: CFGNode.Constant,
 ): List<CFGNode> {
-    val registerArguments = arguments.zip(REGISTER_ARGUMENT_ORDER)
-    val stackArguments = arguments.drop(registerArguments.size).map { Pair(it, Register.VirtualRegister()) }
+    // Argument past standard arguments are assumed to not be references to heap (i.e. static link).
+    val isReference = argTypes.map { it is BaseType.Referential } + List(arguments.size - argTypes.size) { false }
```
AHGIJMKLKKZNPJKQR, line 58: **question:** I don't think I understand what's going on here. I don't think you can have a call with more arguments than the function declares

src/main/kotlin/cacophony/controlflow/functions/PrologueEpilogueHandler.kt
```diff
@@ -10,28 +11,31 @@ class PrologueEpilogueHandler(
...
+        nodes.add(registerUse(rsp, false) subeq stackSpace)
 
-        // Preserved registers
+        // Preserved registers don't hold references
         for ((source, destination) in callConvention.preservedRegisters() zip spaceForPreservedRegisters) {
-            nodes.add(registerUse(destination) assign registerUse(Register.FixedRegister(source)))
+            nodes.add(registerUse(destination, false) assign registerUse(Register.FixedRegister(source), false))
         }
 
+//        val isReference = argTypes.map { it is BaseType.Referential } + List(arguments.size - argTypes.size) { false }
```
AHGIJMKLKKZNPJKQR, line 31: **nitpick:** remove commented out code

src/main/kotlin/cacophony/controlflow/generation/OperatorHandler.kt
```diff
@@ -115,7 +115,7 @@ internal class OperatorHandler(
             is SubCFG.Extracted -> {
                 val assignmentVertex = cfg.addUnconditionalVertex(operatorNode)
                 rhs.exit.connect(assignmentVertex.label)
-                SubCFG.Extracted(rhs.entry, assignmentVertex, noOpOr(SimpleLayout(lhsAccess.access), mode))
+                SubCFG.Extracted(rhs.entry, assignmentVertex, noOpOr(SimpleLayout(lhsAccess.access, false), mode))
```
AHGIJMKLKKZNPJKQR, line 118: **question:** Shouldn't this be true in cases like `@ref += 2`?

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
Giselus: I can, because foreignFunction has to provide the type.

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
Giselus: Truly it's only for static link (it's not on the argument list in the definition, but is passed as regular argument later). Will change to just +1

src/main/kotlin/cacophony/codegen/registers/SpillHandling.kt
Giselus: Wanted to add interference, but I forgor

src/main/kotlin/cacophony/controlflow/generation/OperatorHandler.kt
Giselus: You're right, fixed (probably)

AHGIJMKLKKZNPJKQR (approved):
rocket

PR #282, opened 2025-01-11
Tests for GC

SebastianSpyrzewski:
Created class `GCExamplesTest` for automated testing of GC on example programs (similar to `IOExamplesTest`).

We basically test if 1) programs do not use too much memory and 2) GC did not free memory that is still used (there are no segfaults).


libcacophony.c
```diff
@@ -45,6 +45,16 @@ void put_mem(long long * ptr, long long val) {
     *ptr = val;
 }
 
+// returns integer from [l, r]
+int randint(long long l, long long r) {
+	long long a = rand();
```
FilipKon13, line 50: I don't like using rand() here - the sequence produced by it is implementation defined and can differ on different machines. Reproducibility takes a hit here

libcacophony.c
```diff
@@ -45,6 +45,16 @@ void put_mem(long long * ptr, long long val) {
...
+	return (a % (r - l + 1) + l);
+}
+
+void cassert(long long b) {
+	if (b == 0) exit(1);
```
FilipKon13, line 55: Add print with something like "assertion failed" here, to make this usable in programs outside of the testing environment

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -36,9 +36,6 @@ fun generateCallFrom(
             if (function.arguments.size != arguments.size) {
                 throw IllegalArgumentException("Wrong argument count")
             }
-            if (result != null && !layoutMatchesType(result, function.returnType)) {
```
FilipKon13, line 39: You can add this back I believe? This does not really solve the core issue and fix of the `layoutMatchesType` is in #287

FilipKon13 (changes requested):
Looks pretty good, nice that you included examples where arguments and return values go on the stack. Only minor requests

libcacophony.c
```diff
@@ -45,6 +45,24 @@ void put_mem(long long * ptr, long long val) {
     *ptr = val;
 }
 
+/* ########### Randomness for test purposes ########### */
+unsigned long long rand_seed = 17;
+unsigned long long rand_multiplier = 16807;
+unsigned long long rand_mod = 2147483647;
+unsigned long long rand_() {
+    return rand_seed = ((rand_seed * rand_multiplier) % rand_mod);
+}
```
FilipKon13, line 54: Add const to `rand_multiplier` and `rand_mod`, add static to `rand_seed` and `rand_`

FilipKon13 (changes requested):
Gw. I guess you want to merge after the implementation of gc is on main, or do you want this to be merged with the gc branch?

SebastianSpyrzewski:
> Gw. I guess you want to merge after the implementation of gc is on main, or do you want this to be merged with the gc branch?

I'm not sure, maybe it is better to merge it with gc branch first? This way GC would be tested before merging to main

examples/gc/multiple-references/program.cac
```diff
@@ -0,0 +1,35 @@
...
+let ptr1 = $struct_128;
+let ptr2 = $struct_128;
+let ptr3 = $struct_128;
+let ptr4 = $struct_128;
+let ptr5 = $struct_128;
+
+let i = 0;
+
+while i < 100000 do (     # we are counting number of references to each segment of memory:
+    let temp1 = ptr1;   # (r1=2), r2=1, r3=1, r4=1, r5=1
```
FilipKon13, line 15: There is a major problem with tests it turns out - many of them do not force reference-type variables to land on the stack (coloring puts them in registers or ignores them). My suggestion to force them to land on the stack would be an dummy function call between the declaration and the last use of the reference variable or use of them in a nested function (function does not need to be called)

FilipKon13 (changes requested):

examples/gc/multiple-references/program.cac
FilipKon13: Unresolving, as some tests probably do not test intended behaviour, e.g. this (simple allocation in function):

```
let f = [a: Int] -> Unit => (
    let x = $a;
    if (@x % 2) == 1 then (@x = @x + 1;);
);

let i : Int = 0;
while (i < 10000000) do (
    f[i];
    i = i + 1;
);
```

produces exactly zero references in memory (besides function frames)

examples/gc/allocation-of-structures/program.cac
```diff
@@ -0,0 +1,18 @@
+let struct_128 = {
+    a=1, b=2, c=3, d=4, e=5, f=6, g=7, h=8,
+    i=9, j=10, k=11, l=12, m=13, n=14, o=15, p=16
+};
+
```
FilipKon13, line 5: Could you make this example slightly larger, so that it passes the memory page size (2^12 B) in gc?

FilipKon13 (changes requested):

PR #283, opened 2025-01-12
Lambda expression parsing

mhorod:


FilipKon13:
Would it be possible to add test for the new functionality? Possibly running only part of the pipeline on (atm incorrect) `.cac` files, up to AST generation? Something like

`let f = if true then [x: Int] -> Int => x + 1 else [x: Int] -> Int => x + 2`

FilipKon13 (approved):

PR #286, opened 2025-01-12
Implement zeroing for outlines

honvex-cool:
Implement the functions for clearing object and stack outlines, in C and asm, respectively
Add the call to the label for `clean_refs` with the dedicated `CFGNode.RawCall` in function prologue

### DIFF
--- a/libcacophony.c
+++ b/libcacophony.c
@@ -60,9 +60,11 @@ asm(
 
 "test %rdx, %rsi                            \n" //     if(outline.blocks[i / 64] & (1 << (i % 64)))
 "jz .if_end                                 \n" //     {
-"neg %rax                                   \n" //
-"movq $0, 0(%rbp, %rax, 8)                  \n" //         stack[i] = 0
-"neg %rax                                   \n" //
+
+"neg %rax                                   \n" //         # rax is negated because the offsets from rbp are negative
+"movq $0, -8(%rbp, %rax, 8)                 \n" //         stack[i] = 0
+"neg %rax                                   \n" //         # rax is negated back
+
 ".if_end:                                   \n" //     }
 
 "inc %rax                                   \n" //     i++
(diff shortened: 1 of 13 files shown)


libcacophony.c
```diff
@@ -51,13 +102,50 @@ void put_mem(long long * ptr, long long val) {
 #define GC_WAIT 10
 #endif
 
+#ifndef OUTLINE_PTR_OFFSET_FROM_OBJECT
```
FilipKon13, line 105: These `define`s do not really make code cleaner imo. A comment explaining how outline looks in memory and what points where would be preferable

libcacophony.c
```diff
@@ -28,6 +28,57 @@ asm(
...
+asm(
+".text                                      \n"
+".p2align 4                                 \n"
+".globl clean_refs                          \n"
+".type  clean_refs, @function               \n"
+
+
+"clean_refs:                                \n"
+
+"push %rax                                  \n" // # preserve registers used as local variables
```
FilipKon13, line 40: Can you push temporary registers here on stack? It looks like you can zero out this memory cells in the loop below

FilipKon13 (changes requested):

libcacophony.c
FilipKon13: Resolved

libcacophony.c
```diff
@@ -47,17 +100,51 @@ void put_mem(long long * ptr, long long val) {
...
+ *      ----- increasing addresses at object site ----->
+ *      [8-byte pointer to outline] [8 bytes of object data] [8 bytes of object data] ...
+ *                |                 ^
+ *                |                 |
+ *                |                 the 8 byte pointer returned from `alloc_object` points here
+ *                |
+ *                V
+ *                [S: 8-byte size of object data] [A: 0th 8-byte block of outline] [B: 1st 8-byte block of outline] ...
+ *                ----- increasing addresses at outline site ----->
+ *
```
FilipKon13, line 115: Nice comment

libcacophony.c
```diff
@@ -28,6 +28,59 @@ asm(
...
+
+
+"clean_refs:                                \n"
+
+"push %rax                                  \n" // # preserve registers used as local variables
+"push %rsi                                  \n" //
+"push %rdx                                  \n" //
+"push %rcx                                  \n" //
+
+"xor %rax, %rax                             \n" // i := 0
```
FilipKon13, line 45: I think there is some issue with conventions - we use this asm function after pushing rbp (reference to the parent stack frame) on the stack, and this cell is marked as 'reference holding' by the outline (as it should be), so by starting at 0, you clean that cell

FilipKon13 (changes requested):

libcacophony.c
honvex-cool: Yeah, sorry, I misunderstood the layout at first, but now I'm starting the loop from one so I think that it should be safe.

honvex-cool:
The current implementation seems to be backwards compatible with all the IO tests (they pass), so I believe that at least there are no obvious mistakes. However, most of the implementation is so low-level, I'm not sure how to properly test it separately from the GC :/

src/main/kotlin/cacophony/codegen/registers/SpillHandling.kt
```diff
@@ -78,7 +78,7 @@ fun adjustLoweredCFGToHandleSpills(
         instruction is CopyInstruction &&
             spills.contains(instruction.copyInto()) &&
             spills.contains(instruction.copyFrom()) &&
-            spillsColoring[instruction.copyInto()] === spillsColoring[instruction.copyFrom()]
+            spillsColoring[instruction.copyInto()] == spillsColoring[instruction.copyFrom()]
```
FilipKon13, line 81: What does this do?

src/main/kotlin/cacophony/controlflow/functions/PrologueEpilogueHandler.kt
```diff
@@ -26,6 +27,11 @@ class PrologueEpilogueHandler(
         nodes.add(registerUse(rbp, false) assign registerUse(rsp, false))
         nodes.add(registerUse(rsp, false) subeq stackSpace)
 
+        nodes.add(pushRegister(rdi, false))
+        nodes.add(registerUse(rdi) assign memoryAccess(registerUse(rbp) add integer(REGISTER_SIZE)))
+        nodes.add(CFGNode.RawCall(BlockLabel.cleanReferences))
+        nodes.add(popRegister(rdi, false))
+
```
FilipKon13, line 34: Remove the whole rdi manipulation and just assume inside clean_refs that outline is at [rbp + 8] to save code in the prologue

FilipKon13 (changes requested):
lgtm, nice work. Left minor request with improvement

src/main/kotlin/cacophony/controlflow/functions/PrologueEpilogueHandler.kt
honvex-cool: Done :D

src/main/kotlin/cacophony/codegen/registers/SpillHandling.kt
honvex-cool: Nothing related to my task, I just got tired of the compiler complaining that `warning: Identity equality for arguments of types 'kotlin.Int?' and 'kotlin.Int?' is prohibited`, so I changed it. From what I understand from the [documentation](https://kotlinlang.org/docs/equality.html) (and passsing tests), using `==` seems perfectly fine

FilipKon13 (approved):

PR #289, opened 2025-01-12
add simple escape analysis

### LINKED ISSUE #278: Add escape analysis
I added a proposition of the interface in the `EscapeAnalysis.kt` file, but feel free to adjust it. Insert this inside pipeline in the appropriate place afterwards.

km1chno:
Resolves #278 

## Overview of logic

So I came up with an algorithm for approximating which variables escape scope of function they are declared in. This is what I came up with:

1. It seems that there are two ways variables can escape - through return expressions and assignments to something in scope above - but we have to be careful, because this can be transitive, e.g.
```
let f = [] -> [] -> Int => (
  let x = 10;
  let g = [] -> Int => x;
  let h = g;
  h                         # <-- x escapes indirectly!
)
```
3. First, we preprocess - collect all variables used in all assignments and return expressions.
4. For each variable, we calculate `usageDepth` - the highest static depth of function to which it can escape. This is done in the following way:
- Initialize `usageDepth[var]` to `staticDepth` of function declaring it,
- For all return statements, take all variables it references, and for each set `usageDepth` to static depth of function in which the return is, -1.
- As the last step, until fixed point is obtained, iterate through all assignments. Let L be the set of variables references in LHS, R in the RHS. For each `var \in R`, set `usageDepth` to `min(usageDepth(L))`.
5. We can now identify closures that escape their scope - these are exactly closures whose variables have `usageDepth` smaller than static depth of function declaring it.
6. We identify as escaping all variables corresponding to escaping closures + all variables used in escaping closures, that are not defined in them.

## Caveats - so what could be done better
- If one field of struct escapes, the whole struct is declared as escaping,
- the algorithm in general could be smarter, we could evaluate **for each scope** what can escape it, e.g. we could evaluate better what can escape a return expression, instead of just assuming that everything referenced in it can escape.

## Things left TODO
- [x] Unit tests for more cases
- [x] Insert escape analysis into pipeline


examples/not_yet_correct/closures/escape_via_assignment.cac
```diff
@@ -0,0 +1,7 @@
+let h = [] -> Int => 5;
+
+let f = [] -> () => (
```
FilipKon13, line 3: I think it should be Unit here, not ()

examples/not_yet_correct/closures/escape_via_assignment_struct.cac
```diff
@@ -0,0 +1,7 @@
+let s = {h = () -> Int => 5, y = 10}
+
+let f = [] -> () => (
```
FilipKon13, line 3: As above

FilipKon13 (approved):
Approving beforehand so that you can merge easily

PR #292, opened 2025-01-13
Sanity Restored?

### LINKED ISSUE #288: Void propagation does not work
The code

```
let f = [x : Bool] -> Int => if x then return 0 else return 1;
```

fails with the following error:
```
Exception in thread "main" java.lang.IllegalArgumentException: Failed requirement.
	at cacophony.controlflow.generation.CFGGenerator.makeVerticesForAssignment(CFGGenerator.kt:98)
	at cacophony.controlflow.generation.CFGGenerator.assignLayoutWithValue$tcs(CFGGenerator.kt:65)
	at cacophony.controlflow.generation.CFGGenerator.extendWithAssignment$tcs(CFGGenerator.kt:450)
        ...
```
The issue is connected with the propagation of VoidType and its layout thought the cfg generation.

mhorod:
should resolve #288 :copium:


FilipKon13 (approved):
Nice, passes nontrivial test, approved

PR #293, opened 2025-01-13
Garbage collector implementation

Giselus:


lib_cacophony/gc/gc.cpp
```diff
@@ -0,0 +1,218 @@
...
+}
+
+std::unordered_map<ll*, ll*> memoryManager::cleanup(std::vector<ll*> &alive_objects) {
+    std::vector<memoryPage> pages_to_process;
+    std::swap(pages_to_process, allocated_pages);
+
+    std::set<ll*> alive_objects_set(alive_objects.begin(), alive_objects.end());
+    std::set<ll*> processed_pages;
+    
+    auto traverse_pages = [&](memoryPage page, std::function<void(ll*)> handler) {
```
FilipKon13, line 58: It is not very important, but it would be nice if you could replace `std::function` with somethink more lightweight (perhaps `auto && handler` with suitable comment?)

FilipKon13 (changes requested):

lib_cacophony/gc/gc.cpp
```diff
@@ -0,0 +1,232 @@
...
+#include <cassert>
+
+#ifndef MEMORY_BLOCK_SIZE
+#define MEMORY_BLOCK_SIZE (1 << 12)
+#endif
+
+void memoryManager::createNewPage(int size) {
+    assert(size % sizeof(ll*) == 0);
+    ll *ptr = static_cast<ll*>(malloc(size));
+    std::memset(ptr, 0, size);
```
FilipKon13, line 16: Use `calloc()` here

lib_cacophony/gc/gc.cpp
```diff
@@ -0,0 +1,232 @@
...
+    auto page = memoryPage{size, 0, ptr};
+    allocated_pages.push_back(page);
+}
+
+ll* memoryManager::allocateMemory(int size) {
+    assert(size % sizeof(ll*) == 0);
+    if (size > MEMORY_BLOCK_SIZE) {
```
FilipKon13, line 23: Make sure that there is a test that covers this

lib_cacophony/gc/gc.cpp
```diff
@@ -0,0 +1,232 @@
+#include "gc.h"
+#include "gcimpl.h"
+
+#include <cstring>
+#include <functional>
```
FilipKon13, line 5: The `<functional>` header can be removed I believe

lib_cacophony/gc/gc.cpp
```diff
@@ -0,0 +1,232 @@
...
+        if(next_is_stack_frame || !isVisited(referenced_object)) {
+            traverseObjects(referenced_object, next_is_stack_frame);
+        }
+    }
+}
+
+void objectTraversal::remapReferences(ll *rbp, std::unordered_map<ll*, ll*> mapping) {
+    clear();
+    remap_refs = true;
+    reference_mapping = mapping;
```
FilipKon13, line 196: Add assignment via `std::move`

lib_cacophony/gc/gc.cpp
```diff
@@ -0,0 +1,232 @@
...
+static ll getObjectSize(ll *outline) {
+    return 8 * (1 + *outline);
+}
+
+static void runGc(ll *rbp) {
+    if(LOG_GC) std::cerr << "running gc" << std::endl;
+    objectTraversal object_traversal;
+    std::set<ll*> alive_objects = object_traversal.getAliveReferences(rbp);
+    std::unordered_map<ll*, ll*> reference_mapping = memory_manager.cleanup(alive_objects); 
+    object_traversal.remapReferences(rbp, reference_mapping);
```
FilipKon13, line 221: `std::move` into function or change argument type to reference (preference for the first option at the current state)

lib_cacophony/gc/gc.h
```diff
@@ -126,18 +84,9 @@ void put_mem(long long * ptr, long long val) {
  */
 
 #ifndef GC_WAIT
-#define GC_WAIT 10
+#define GC_WAIT 1
```
FilipKon13, line 87: : O

lib_cacophony/gc/gcimpl.h
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+  public:
+    void remapReferences(ll *rbp, std::unordered_map<ll*, ll*> mapping);
+    std::set<ll*> getAliveReferences(ll *rbp);
+};
+
+static void runGc(ll *rbp);
+
+static ll** allocMemory(ll *outline);
```
FilipKon13, line 57: Remove static functions declarations from header files

FilipKon13 (changes requested):

lib_cacophony/gc/gc.h
Rafi0221: for testing ;D

lib_cacophony/gc/gcimpl.h
Rafi0221: Can't remove this one, it is used in `gc.h`

FilipKon13: gc.h does not contain any declarations atm, everything has implementation and could be moved to gc.cpp (or separate source file)

Rafi0221: Yes, but I wanted to separate out internal implementation with functions available in the compiler. Maybe after the merge it is not as clear

FilipKon13 (approved):

Giselus:
image

PR #294, opened 2025-01-14
Change to unconditional zeroing

honvex-cool:
Asm reference cleaning changed to a simple loop to write zeros on the stack
C reference cleaning changed to `memset`


Giselus (approved):
LGTM 👍 You can merge

PR #303, opened 2025-01-19
Add example programs using functional features

### LINKED ISSUE #290: Add lambda expression examples
In #283 we added tests that check partial correctness (AST generation) of programs using lambda expressions.
Refactor them into actual correct examples, and add some new ones.

km1chno:
Resolves #290 

I'm not sure whether some examples in `not_yet_correct` should actually just be `incorrect`, as I know that our scope for functional features is quite dynamic at the moment:P So maybe some of them will have to be moved to incorrect examples.

There are 3 tests in `not_yet_correct` which are IO tests, but obviously they have not been run yet

## Cases covered

- Obtaining functions/lambdas as arguments
- Returning functions/lambdas as function value
- Assignment of lambdas to named functions
- Some more convoluted variable escaping through struct fields and transitivity


mhorod (approved):
Looks great!

I don't know either what will be the future of some of the examples, but I guess we will discuss this during next labs.

PR #304, opened 2025-01-19
Closure analysis and outline

Rafi0221:


src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
```diff
@@ -189,6 +187,13 @@ class CacophonyPipeline(
             }.filterIsInstance<Definition.ForeignFunctionDeclaration>()
             .toSet()
 
+    private fun analyzeClosures(escapeAnalysis: EscapeAnalysisResult /*, lambdaAnalysis: LambdaAnalysisResult */): ClosureAnalysisResult {
```
mhorod, line 190: I'm not sure about the commenting out the code.
Since we have this noted in TODOs, I'd either remove them, or uncomment them and add some sketch of the interface we'd like here (and all other places that are commented out).

src/main/kotlin/cacophony/semantic/analysis/ClosureAnalysis.kt
```diff
@@ -8,6 +8,9 @@ import cacophony.semantic.syntaxtree.LambdaExpression
  */
 typealias ClosureAnalysisResult = Map<LambdaExpression, Set<Variable>>
 
-fun analyseClosures(): ClosureAnalysisResult {
-    return emptyMap() // TODO - implement closure analysis
+fun analyseClosures(escapeAnalysis: EscapeAnalysisResult /*, lambdaAnalysis: LambdaAnalysisResult */): ClosureAnalysisResult {
+//    return lambdaAnalysis.mapValues { (_, analysis) ->
```
mhorod, line 12: Like above, if this actually works under some sensible interface of `LambdaAnalysisResult` then I think we can have this uncommented.

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
```diff
@@ -251,6 +253,8 @@ class LambdaExpression(
     FunctionalExpression {
     override fun toString() = "[${arguments.joinToString(", ")}] -> $returnType"
 
+    override fun getLabel() = "Lambda_${arguments.size}_${hashCode().toULong()}"
```
mhorod, line 256: This feels a bit hacky, and I'm not sure if this is going to be consistent between runs (and if it's not then snapshots will fail, no?)
but I guess we can leave it if it works.

mhorod (approved):
Overall looks great, I'm not sure about some little things I pointed out, but I think this is quite enough for this sprint.

src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt
Rafi0221: That's how we generate all labels and it works great
https://github.com/mhorod/cacophony/blob/7a6f7bb6e2d437ea5178b49c412461da893fc2f9/src/main/kotlin/cacophony/semantic/syntaxtree/SyntaxTree.kt#L214-L218

src/main/kotlin/cacophony/pipeline/CacophonyPipeline.kt
Rafi0221: We decided to leave it commented out

Issue #305, opened 2025-01-19
Lambda implementation TODO


mhorod:
List of issues found during the implementation of lambdas.

Rafi0221:
* Fix `escapeAnalysis` to use `lambdaAnalysis` along/instead of `functionAnalysis`
* Uncomment the implementation of `closureAnalysis`
* Implement `generateLambdaHandlers`
* Adjust `analyzedFunction` to analyze lambdas
* Adjust `ObjectOutlinesCreator` to correctly generete the outline for `FunctionType`

FilipKon13:
Closure analysis is not really used in the `FunctionHandlerImpl`, as it does not need to generate a closure

SebastianSpyrzewski:
- Heap variables need to be allocated somewhere, I was thinking about prologue - it gets map from heap variables to their pointers (as PrimitiveVariables) as argument
- `PrologueEpilogueHandler` line 28 won't work for `LambdaHandler`
- can named function become closures? `LambdaHandler` is defined only for `LambdaExpression` now
- someone needs to create closure for lambda

- maybe not necessary, but it would be nice to not use AST nodes in handlers anymore, all the info like arguments and return type could be in `AnalyzedFunction`

SebastianSpyrzewski:
Tidy up tests for `FunctionHandler`, `LambdaHandler`

PR #306, opened 2025-01-19
Create lambda handlers

SebastianSpyrzewski:
Created lambda handler and made it access correctly variables from closure.

Some TODO's in PrologueEpilogueHandler and FunctionAnalysis that I believe need to be done next week.


src/main/kotlin/cacophony/controlflow/CFGUtils.kt
```diff
@@ -90,5 +90,9 @@ fun wrapAllocation(allocation: VariableAllocation, holdsReference: Boolean): CFG
     when (allocation) {
         is VariableAllocation.InRegister -> registerUse(allocation.register, holdsReference)
         is VariableAllocation.OnStack -> memoryAccess(registerUse(rbp, false) sub integer(allocation.offset), holdsReference)
-        is VariableAllocation.ViaPointer -> TODO()
+        is VariableAllocation.ViaPointer ->
+            memoryAccess( // should holdsReference be set as true? I'm not sure
```
mhorod, line 94: I'm not sure either, better ask the author of this field 😄

src/main/kotlin/cacophony/controlflow/functions/FunctionHandlerImpl.kt
```diff
@@ -1,88 +1,42 @@
...
-import cacophony.semantic.syntaxtree.Definition
 import cacophony.semantic.syntaxtree.Definition.FunctionDefinition
-import kotlin.math.max
 
 class FunctionHandlerImpl(
     private val function: FunctionDefinition,
     private val analyzedFunction: AnalyzedFunction,
     // List of parents' handlers ordered from immediate parent.
-    private val ancestorFunctionHandlers: List<FunctionHandler>,
+    private val ancestorFunctionHandlers: List<CallableHandler>, // first (or last?) can be Lambda, not Function
```
mhorod, line 15: does this matter? 🤔

SebastianSpyrzewski: wdym?

mhorod: I mean, I don't really understand what's this comment for 😄 
Does it mean that we have to be careful what we provide in the constructor? Or how we use these handlers? Or is it just explaning why it's `CallableHandler` and not `FunctionHandler`.

If it's the last one then I think we don't have to explain it, it feels natural that the interface is as general as it can be.

src/main/kotlin/cacophony/controlflow/functions/CallableHandlerImpl.kt
```diff
@@ -0,0 +1,162 @@
...
+import cacophony.controlflow.generation.Layout
+import cacophony.controlflow.generation.generateLayoutOfVirtualRegisters
+import cacophony.semantic.analysis.AnalyzedFunction
+import cacophony.semantic.analysis.EscapeAnalysisResult
+import cacophony.semantic.analysis.VariablesMap
+import cacophony.semantic.syntaxtree.FunctionalExpression
+import kotlin.math.max
+
+abstract class CallableHandlerImpl(
+    private val analyzedFunction: AnalyzedFunction, // TODO: we need to change it so it also covers lambda expressions
```
mhorod, line 15: You can also describe the TODOs here so it's all in one place: https://github.com/mhorod/cacophony/issues/305

mhorod (approved):
Looks great, given that this was a bit more difficult sprint due to the confusion, good work!

PR #307, opened 2025-01-19
Via-Pointer variable allocation

FilipKon13:

PR #309, opened 2025-01-20
Dont mark escaping variables via assignments to non-functional types

km1chno:
## Changes

- don't mark escaping variables via assignments to non-functional types
- add `--log-escape-analysis` flag to compiler and appropriate functions to logger


km1chno:
Merged as part of #307

PR #311, opened 2025-01-20
Fix let x = x + 1

km1chno:
Changes name resolver AST traversal so that value of variable declaration is visited before the new identifier is "created" - so that it is not visible while traversing value expression.


Rafi0221 (approved):
I confirm that this change resolves the issue🐱

PR #312, opened 2025-01-20
Add lambda cfg gen

AHGIJMKLKKZNPJKQR:


src/main/kotlin/cacophony/codegen/instructions/matching/InstructionMatcher.kt
```diff
@@ -15,15 +15,14 @@ interface InstructionMatcher {
...
 
-fun slotFillFromMetadata(valueMapping: ValueSlotMapping, metadata: InstructionMatcherImpl.MatchMetadata): SlotFill {
-    return SlotFill(
+fun slotFillFromMetadata(valueMapping: ValueSlotMapping, metadata: InstructionMatcherImpl.MatchMetadata): SlotFill =
+    SlotFill(
         valueMapping,
         metadata.registerFill,
         metadata.constantFill,
-        metadata.functionFill,
+        // metadata.functionFill,
```
mhorod, line 23: If we don't need that, why not remove this at all?

src/main/kotlin/cacophony/controlflow/functions/GenerateCall.kt
```diff
@@ -120,7 +86,8 @@ fun generateCall(
         nodes.add(pushRegister(register, register.holdsReference))
     }
 
-    nodes.add(call(function))
+    // Tomek: "Chyba będzie działało :>"
```
mhorod, line 89: XD

src/main/kotlin/cacophony/controlflow/generation/Layout.kt
```diff
@@ -1,25 +1,21 @@
...
 class SimpleLayout(val access: CFGNode, val holdsReference: Boolean = false) : Layout() {
     override fun flatten(): List<LayoutAccessInfo> = listOf(LayoutAccessInfo(access, holdsReference))
+}
 
-    override fun matchesType(type: Type): Boolean =
-        when (type) {
-            is BaseType.Basic, is BaseType.Referential -> true
-            is BaseType.Functional, is BaseType.Structural -> false
-        }
+// NOTE: link is either the static link or closure link
```
mhorod, line 16: You can convert this to a javadoc `/** ... */` with description of the fields.

mhorod (approved):
Looks good, nice work 🫡 

Just a small suggestion - I see there's a lot of code that is commented out, and I think that we can get rid of most of it (that's what git is for), and it would be awesome if you could also describe the TODOs and issues here, so we have a clearer view from above: https://github.com/mhorod/cacophony/issues/305

src/main/kotlin/cacophony/codegen/instructions/matching/InstructionMatcher.kt
jjaworska: Yeah, this was just laziness :laughing:  . On it

PR #313, opened 2025-01-20
Add memory limits to gc tests

Giselus:


lib_cacophony/gc/gc.cpp
```diff
@@ -19,6 +19,17 @@ void memoryManager::createNewPage(int size) {
     ll *ptr = static_cast<ll*>(malloc(size));
     auto page = memoryPage{size, 0, ptr};
     allocated_pages.push_back(page);
+    occupied_memory += size;
+}
+
+// This step is performed during cleanup, where allocated_pages is being modified, therefore no need to modify it heee.
```
mhorod, line 25:
```suggestion
// This step is performed during cleanup, where allocated_pages is being modified, therefore no need to modify it here.
```

mhorod (approved):
Looks great, nice optimization of the page traversal :)

PR #315, opened 2025-01-21
FunctionDefinition and FunctionDeclaration are no more

mhorod:


mhorod:
TODO:
- [ ]  Escape analysis compilation still fails:
  - `visitFunctionDeclaration` has to be removed/replaced since we removed `FunctionDeclaration`
  - For some reason we get functions from `functionAnalysis` and then lookup them in `variableMap`, but now lambda expressions are not variables so it doesn't typecheck
- [ ] `NameResolver.define` has to be adjusted for new interface
- [ ] `OverloadResolver` returned map from `Int` to `FunctionDeclaration` but now `FunctionDeclaration` is gone, so what?
- [ ] `OverloadResolver` returns map from `VariableUse` to `Definition`, so it probably needs a map from `LambdaExpression` to `Definition` for static functions?
- [ ] `TypeChecker` might want to access declared type of typed `LambdaExpression`
- [ ] remove todos about `FunctionalExpression` since we won't use it?

mhorod:
conclusions after przodownik meeting:
- We can remove call graph and var use type analysis
- FunctionAnalysis should be done for all LambdaExpressions, and it can ignore distinction between lambdas and static functions (task for @Rafi0221):
   - ignore calls and use only syntactic relations of using a variable inside the body
 -   Closure analysis might be redundant
 - Remove SideEffectAnalyzer from CFG and always evaluate from left to right

mhorod:
State for now:
- Removed `FunctionDeclaration` which was the common interface of `FunctionDefinition` and `ForeignFunctionDeclaration`
- Almost everything that previously operated on `FunctionDefinition` now operates on `LambdaExpression`
- We added back `FunctionDefinition` ast node, but it's subclass of `VariableDeclaration` and it's not as widely used as before
- We added `NamedFunctionInfo` which maps `LambdaExpression` to its `FunctionDefinition`, this is necessary for making a final distinction which functions will be closures and which ones should be static
- We removed `CallGraph`, `VarUseTypeAnalysis` and `SideEffectAnalyzer` - these three were ultimately used only to determine the evaluation order in binary operators and blocks. This is not really a crucial feature as we can always evaluate  left to right, and we don't have to worry about fixing this code and adjusting it for lambdas which seems nontrivial.
- Consequently, `FunctionAnalysis` has to be adjusted a bit - we need it to return information about used variables that is later used for allocation and escape analysis, and it should be done for all `LambdaExpression`s  without distinction between static and closure functions. 
- `ClosureAnalysis` is a bit repurposed - since `FunctionAnalysis` and `EscapeAnalysis` are responsible for finding captured and escaping variables, `ClosureAnalysis` now decides which functions are static and which are closures. This is then used to generate handlers.
- We added `CallableHandlers` that collect all handlers for closures and static functions so we can store them in one object instead of two.

TODO:
- Implement `generateCallableHandlers` - it should look which functions are closures/static and call appropriate function
- Fix commented code `visitAssignable` in CFGGenerator
- Log new closure analysis in `CacophonyLogger`
- Implement missing parts of `EscapeAnalysis`, think if it really needs distinction between (syntactically) static and closures functions
- Adust `analyzeFunctions` so it works for all functions
- Implement `getNamedFunctions`
- Fix commented out code in `NameResolver`
- Fix tests, at least compilation, since they will probably fail in runtime due to changes.

src/main/kotlin/cacophony/codegen/SafeLinearization.kt
```diff
@@ -100,7 +100,7 @@ private fun handleSpills(
                 functionDeclaration to
```
SebastianSpyrzewski, line 100: Why is it still called functionDEclaration if it is LambdaExpression now?

src/main/kotlin/cacophony/codegen/SafeLinearization.kt
```diff
@@ -100,7 +100,7 @@ private fun handleSpills(
...
                     adjustLoweredCFGToHandleSpills(
                         instructionCovering,
-                        functionHandlers[functionDeclaration]!!,
+                        staticFunctionHandlers[functionDeclaration]!!,
```
SebastianSpyrzewski, line 103: And why do we assume it has static handler, when it can be closure as well?

src/main/kotlin/cacophony/controlflow/functions/GenerateHandlers.kt
```diff
@@ -51,7 +64,23 @@ fun generateFunctionHandlers(
...
+        handlers[lambda] =
+            ClosureHandlerImpl(
+                callConvention,
+                lambda,
+                analyzedFunctions[lambda]!!,
+                variablesMap,
+                escapeAnalysisResult,
+            )
+    }
+    return handlers
```
honvex-cool, line 85: NIT: I believe that it would be more idiomatic to use `closureCallables.map`.

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -533,6 +560,10 @@ internal class CFGGenerator(
         return SubCFG.Extracted(vertex, artificialExit, VoidLayout())
     }
 
+    /*
+        let f = [] => ()
+     */
+
```
honvex-cool, line 566: This comment seems out of place. Could you please elaborate on it or delete it?

src/main/kotlin/cacophony/controlflow/generation/OperatorHandler.kt
```diff
@@ -21,13 +20,7 @@ internal class OperatorHandler(
...
         val rhsCFG = cfgGenerator.visit(expression.rhs, mode, context)
 
-        val safeLhs =
-            if (sideEffectAnalyzer.hasClashingSideEffects(expression.lhs, expression.rhs)) {
-                // If there are clashing side effects, lhs must be extracted to a separate vertex
-                cfgGenerator.ensureExtracted(lhsCFG, mode)
-            } else {
-                lhsCFG
-            }
+        val safeLhs = cfgGenerator.ensureExtracted(lhsCFG, mode)
```
honvex-cool, line 23: Maybe we could keep this logic and the `SideEffectAnalyzer` in CFG generation? For now, just make `SideEffectAnalyzer` an interface and use an implementation that always returns `true` (there *are* clashing side effects).
If there is some time left, we could implement simpler variable use analysis which does not use the call graph (because, from what I understand, the call graph is the root of all evil). Then, we could switch the `SideEffectAnalyzer` implementation and restore the feature that we used to have - in my opinion, it's good that we can put simple expressions like `x + y` into a single CFG vertex when there are no clashing side effects.

src/main/kotlin/cacophony/semantic/names/NameResolver.kt
```diff
@@ -6,9 +6,8 @@ import cacophony.diagnostics.NRDiagnostics
 import cacophony.semantic.syntaxtree.*
 import cacophony.semantic.syntaxtree.Definition.ForeignFunctionDeclaration
 import cacophony.semantic.syntaxtree.Definition.FunctionArgument
-import cacophony.semantic.syntaxtree.Definition.FunctionDeclaration
-import cacophony.semantic.syntaxtree.Definition.FunctionDefinition
 import cacophony.semantic.syntaxtree.Definition.VariableDeclaration
+import cacophony.semantic.syntaxtree.LambdaExpression
 import cacophony.utils.CompileException
 
 class NameResolutionException(
```
honvex-cool, line 13: This is unsed, might as well get rid of it now.

src/main/kotlin/cacophony/semantic/types/TypeChecker.kt
```diff
@@ -52,28 +52,11 @@ private class Typer(
...
-                    BuiltinType.UnitType
-                }
-
                 is LambdaExpression -> {
                     val argsType = parseArgs(expression.arguments) ?: return null
                     val returnType = translator.translateType(expression.returnType) ?: return null
-                    val deducedType = FunctionType(argsType, returnType)
-                    val functionType = initializedType(null, deducedType, expression.range) ?: return null
+//                    val deducedType = FunctionType(argsType, returnType)
+//                    val functionType = initializedType(null, deducedType, expression.range) ?: return null
```
SebastianSpyrzewski, line 59: why is this commented out? If it is not useful, maybe we can just delete it

SebastianSpyrzewski (changes requested):
I see an issue in spill handling, and few different minor things.

Generally looks like a good refactor, nice job!

(But I must admit I did not go through all the tests)

honvex-cool (commented):
Wow, I admire your work. This changes look painful to implement. I left some minor comments.

src/main/kotlin/cacophony/controlflow/functions/GenerateHandlers.kt
jjaworska: I think `map` wasn't the fit here, but I used `associateWith`

src/main/kotlin/cacophony/controlflow/functions/CallableHandlers.kt
```diff
@@ -0,0 +1,30 @@
...
+    val staticFunctionHandlers: Map<LambdaExpression, StaticFunctionHandler>,
+) {
+    fun getCallableHandler(callable: LambdaExpression): CallableHandler =
+        if (closureHandlers.containsKey(callable)) {
+            closureHandlers[callable] ?: error("value was just in map")
+        } else if (staticFunctionHandlers.containsKey(callable)) {
+            staticFunctionHandlers[callable] ?: error("value was just in map")
+        } else {
+            error("No handler found for callable $callable")
+        }
```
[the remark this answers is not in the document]
jjaworska: I like this, thanks!

src/main/kotlin/cacophony/controlflow/functions/CallableHandlerImpl.kt
```diff
@@ -6,17 +6,16 @@ import cacophony.controlflow.CFGNode.RegisterUse
...
 
-abstract class CallableHandlerImpl(
-    private val analyzedFunction: AnalyzedFunction, // TODO: we need to change it so it also covers lambda expressions
-    private val function: FunctionalExpression, // TODO: and probably get rid of that, analyzedFunction should have all necessary info
+sealed class CallableHandlerImpl(
+    private val analyzedFunction: AnalyzedFunction,
+    private val function: LambdaExpression,
     private val variablesMap: VariablesMap,
-    private val escapeAnalysisResult: EscapeAnalysisResult,
+    private val escapedVariables: Set<Variable>,
```
Giselus, line 18: Why are we using Set<Variable> here and EscapeAnalysisResult in other places? Isn't it the same thing?

src/main/kotlin/cacophony/controlflow/functions/ClosureHandlerImpl.kt
```diff
@@ -4,19 +4,16 @@ import cacophony.controlflow.*
...
+class ClosureHandlerImpl(
     callConvention: CallConvention,
-    private val function: LambdaExpression, // TODO: it can be FunctionalExpression
+    private val function: LambdaExpression,
     private val analyzedFunction: AnalyzedFunction,
     private val variablesMap: VariablesMap,
-    escapeAnalysisResult: EscapeAnalysisResult,
-    closureAnalysisResult: ClosureAnalysisResult,
-) : LambdaHandler, CallableHandlerImpl(
+    escapeAnalysisResult: Set<Variable>,
```
Giselus, line 15: Why type changed from EscapeAnalysisResult to Set<Variable> if EscapeAnalysisResult is an alias for Set<Variable>?

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
```diff
@@ -556,11 +587,7 @@ internal class CFGGenerator(
 
     private fun getCurrentCallableHandler(): CallableHandler = getCallableHandler(function)
 
-    private fun getCallableHandler(callable: FunctionalExpression): CallableHandler =
-        when (callable) {
-            is Definition.FunctionDefinition -> functionHandlers[function] ?: error("Function $function has no handler")
-            is LambdaExpression -> lambdaHandlers[function] ?: error("Lambda $function has no handler")
-        }
+    private fun getCallableHandler(callable: LambdaExpression): CallableHandler = callableHandlers.getCallableHandler(callable)
```
Giselus, line 590: Do we need this wrapper? It's used only in one place.

src/main/kotlin/cacophony/semantic/analysis/NamedFunctionInfo.kt
```diff
@@ -0,0 +1,61 @@
+package cacophony.semantic.analysis
+
+import cacophony.semantic.syntaxtree.*
+
+/**
+ * For each lambda expression that was defined as part of syntactically static function store this function
+ */
+typealias NamedFunctionInfo = Map<LambdaExpression, Definition.FunctionDefinition>
+
+private class GetNamedFunctionsVisitor {
```
Giselus, line 10: Maybe just _NamedFunctionsVisitor_

src/main/kotlin/cacophony/semantic/analysis/NamedFunctionInfo.kt
```diff
@@ -0,0 +1,61 @@
...
+    private val accum: MutableMap<LambdaExpression, Definition.FunctionDefinition> = mutableMapOf()
```
Giselus, line 11: It's clear what that is, but maybe we can use more descriptive name, at least to be consistent with other visitors in our code.

Giselus (changes requested):
Great job! A lot of work has been done here. I see that most of the important issues have already been caught, so I've only left a few cosmetic suggestions

src/main/kotlin/cacophony/controlflow/functions/ClosureHandlerImpl.kt
Giselus: It's the same as previous comment, and I believe it was like that in few other places, but I didn't want to spam with marking all of them.

src/main/kotlin/cacophony/controlflow/functions/GenerateHandlers.kt
```diff
@@ -1,33 +1,46 @@
...
+    staticLinkCallables: Set<LambdaExpression>,
+    callConvention: CallConvention,
+    variablesMap: VariablesMap,
+    escapedVariables: Set<Variable>,
+): Map<LambdaExpression, StaticFunctionHandler> {
+    val handlers = mutableMapOf<LambdaExpression, StaticFunctionHandler>()
+    val order = analyzedFunctions.filter { staticLinkCallables.contains(it.key) }.entries.sortedBy { it.value.staticDepth }
 
-    val ancestorHandlers = mutableMapOf<Definition.FunctionDefinition, List<FunctionHandler>>()
+    val ancestorHandlers = mutableMapOf<LambdaExpression, List<StaticFunctionHandler>>()
```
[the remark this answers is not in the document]
mhorod: yup, good catch

src/main/kotlin/cacophony/controlflow/generation/OperatorHandler.kt
mhorod: That's a good point, although we probably won't develop it further, it looks a bit nicer.

src/main/kotlin/cacophony/controlflow/functions/CallableHandlerImpl.kt
mhorod: It wasn't for a short while during refactor, but now you're right 😄

src/main/kotlin/cacophony/controlflow/generation/CFGGenerator.kt
mhorod: good point, inlined this method

SebastianSpyrzewski (approved):

Giselus (approved):

src/main/kotlin/cacophony/controlflow/functions/CallableHandlerImpl.kt
```diff
@@ -9,14 +9,14 @@ import cacophony.semantic.analysis.AnalyzedFunction
...
 
-abstract class CallableHandlerImpl(
-    private val analyzedFunction: AnalyzedFunction, // TODO: we need to change it so it also covers lambda expressions
-    private val function: FunctionalExpression, // TODO: and probably get rid of that, analyzedFunction should have all necessary info
+sealed class CallableHandlerImpl(
+    private val analyzedFunction: AnalyzedFunction,
+    private val function: LambdaExpression,
     private val variablesMap: VariablesMap,
-    private val escapeAnalysisResult: EscapeAnalysisResult,
+    private val escapedVariables: EscapeAnalysisResult
```
SebastianSpyrzewski, line 19: ktlint is crying about missing comma here

mhorod: fixed

PR #323, opened 2025-01-25
Add FunctionVariable and visit lhs in function call

AHGIJMKLKKZNPJKQR:
Variables are created for everything, i.e.:
- For "normal" variables i.e. primitives and structs
- For anonymous functions `let x = ([] -> T => E)`
- For named functions `let f = [] -> T => E`


src/main/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysis.kt
```diff
@@ -74,16 +75,7 @@ private class VarUseVisitor(
...
-        val result = HashSet<Variable>()
-        result.add(variable)
-        if (variable is Variable.StructVariable) {
-            variable.fields.values.forEach {
-                result.addAll(gatherNestedVariables(it))
-            }
-        }
-        return result
-    }
+    private fun gatherNestedVariables(variable: Variable) = getAllNestedVariables(variable).toSet()
```
SebastianSpyrzewski, line 78: I think function `gatherNestedVariables` can be completely removed now, it is just unnecessary wrap

SebastianSpyrzewski (commented):
Looking good, is something yet to be done here besides merging with other PRs?

src/main/kotlin/cacophony/semantic/analysis/VarUseTypeAnalysis.kt
AHGIJMKLKKZNPJKQR: this file is gone, so resolving the comment

Giselus (approved):
Great 👍

PR #325, opened 2025-01-26
Fix FunctionAnalysis and StaticFunctionRelations

Rafi0221:


src/main/kotlin/cacophony/semantic/analysis/StaticFunctionRelations.kt
```diff
@@ -79,19 +77,19 @@ private class StaticFunctionsRelationsVisitor(
         functionStack.lastOrNull()?.let {
             relations[it]?.usedVariables?.add(UsedVariable(variable, useType))
         }
-        if (variable is Variable.StructVariable) {
-            variable.fields.forEach {
-                markNestedVariables(it.value, useType)
-            }
+        if (variable !is Variable.FunctionVariable) {
+            // Primitives inside FunctionVariable are not variables present in the compiled code
```
SebastianSpyrzewski, line 81: I don't think this is fine, handler is operating on `PrimitiveVariable`s, so maybe it will need information about primitives corresponding to code pointer and static/closure link.

In particular, in `CallableHandlerImpl` line 74 we are doing `getPrimitives()`, which is fine with this code, as `getPrimitives()` on function variable will give us both of these variables, but in line 69 we are using `filterIsInstance<Variable.PrimitiveVariable>`, which will lose the information about them, as they are not present in corresponding set.

We can make this work, but IMO this is room for possible bugs and I don't see what are we gaining by not marking these Variables as used.

Also, comment is a bit confusing - what do you mean by "are not present"? When we do assignments on function variables we need to assign both of these variables.

Rafi0221: I think I agree with you. Removed the if.

Giselus (approved):
Looks okay I think, but please, fix linter errors 🥺

PR #326, opened 2025-01-27
Add generation of callable handlers

honvex-cool:


src/main/kotlin/cacophony/semantic/analysis/FunctionIdentities.kt
```diff
@@ -29,7 +33,10 @@ private class NamedFunctionVisitor {
                 visit(ast.function)
                 ast.arguments.forEach { visit(it) }
             }
-            is LambdaExpression -> visit(ast.body)
+            is LambdaExpression -> {
+                anonymousFunctions.add(ast)
```
SebastianSpyrzewski, line 37: this way for `LambdaExpression` being part of FunctionDefinition it will be in both `namedFunctions` and `anonymousFunctions`. Is that an intended behavior? Name of those collections suggests otherwise.

If we want to avoid that, we can maybe implement it like
```
is Definition.FunctionDefinition -> {
                namedFunctions[ast.value] = ast
                visit(ast.value.body)
            }
```

src/main/kotlin/cacophony/controlflow/functions/GenerateHandlers.kt
```diff
@@ -14,69 +13,42 @@ fun generateCallableHandlers(
...
-                StaticFunctionHandlerImpl(
-                    function,
-                    analyzedFunction,
-                    emptyList(),
-                    callConvention,
-                    variablesMap,
-                    escapeAnalysis,
-                )
+    analyzedFunctions.entries.sortedBy { it.value.staticDepth }.forEach { (lambda, analyzed) ->
+        if (closureAnalysis.staticFunctions.contains(lambda)) {
```
SebastianSpyrzewski, line 25: This `if` should be the other way around, you are creating ClosureHandler for static functions and StaticHandler for closures 😉

SebastianSpyrzewski (changes requested):
Generally looks good, but I gave some comments.

Please add some simple tests for `ClosureAnalysis` and `GenerateHandlers` (there were no tests for latter before, but now there is some logic here, so it should get tested) and fix linter errors.

src/main/kotlin/cacophony/controlflow/functions/GenerateHandlers.kt
honvex-cool: Done :D

PR #329, opened 2025-01-27
Better overload resolution

MisieqqeisiM:


src/main/kotlin/cacophony/diagnostics/TypeCheckerDiagnostics.kt
```diff
@@ -62,4 +62,12 @@ class TypeCheckerDiagnostics {
     ) : DiagnosticMessage {
         override fun getMessage() = "Cannot allocate value of type $type"
     }
+
+    data object TooManyOverloads : DiagnosticMessage {
+        override fun getMessage() = "Too many possible overloads in assignment"
+    }
+
+    data object TooFewOverloads : DiagnosticMessage {
+        override fun getMessage() = "Non of the overloads match"
```
SebastianSpyrzewski, line 71:
```suggestion
        override fun getMessage() = "None of the overloads match"
```

src/main/kotlin/cacophony/semantic/names/NameResolver.kt
```diff
@@ -3,246 +3,307 @@ package cacophony.semantic.names
...
+
+    private fun decideSingleShape(annotation: Type?, value: Expression): Shape {
+        val expectedShape = annotation?.let { Shape.from(it) } ?: Shape.Top
+        val validShapes = traverseAst(value, true).filter { it isSubshapeOf expectedShape }
+        val decidedShape =
+            if (validShapes.size == 1) {
+                val shape = validShapes.last()
+                if (annotation == null) shape else expectedShape
+            } else {
+                expectedShape
```
Giselus, line 214: Don't we have to check if expectedShape is inside validShapes?

src/main/kotlin/cacophony/semantic/names/NameResolver.kt
```diff
@@ -3,246 +3,307 @@ package cacophony.semantic.names
...
+        blocks.lastOrNull()?.let { block ->
+            block.keys.forEach { id -> idToBlocks[id]!!.removeLast() }
+            blocks.removeLast()
+        }
+    }
 
-            is LambdaExpression -> {
-                visitLambdaExpression(node.arguments, node.body)
-            }
+    fun define(id: String, definition: Definition, arity: Int? = null) {
```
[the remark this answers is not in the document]
MisieqqeisiM: it's a relic from a previous version when lambdas were supposed to be overloadable

src/main/kotlin/cacophony/semantic/names/NameResolver.kt
MisieqqeisiM: it's not necessary since shapes are only hints for picking an appropriate overload in the typechecker. The compilation will fail later on.

Giselus (approved):
I just have to believe that this works

PR #332, opened 2025-01-27
Alloc ViaPointer Variables

FilipKon13:
TODO: add tests and deal with the case when function argument has ViaPointer allocation (afaik, it can happen?)

### DIFF
--- a/src/test/kotlin/cacophony/controlflow/generation/FunctionPrologueAndEpilogueTest.kt
+++ b/src/test/kotlin/cacophony/controlflow/generation/FunctionPrologueAndEpilogueTest.kt
@@ -3,9 +3,6 @@ package cacophony.controlflow.generation
 import cacophony.*
 import cacophony.codegen.BlockLabel
 import cacophony.controlflow.*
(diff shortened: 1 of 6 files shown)


Giselus (approved):
Prolly good, would be nice if we could test it somehow tho

src/test/kotlin/cacophony/controlflow/generation/FunctionPrologueAndEpilogueTest.kt
```diff
@@ -179,12 +182,118 @@ class FunctionPrologueAndEpilogueTest {
...
+         */
+        val xDef = variableDeclaration("x", lit(2))
+        val fDef = unitFunctionDefinition("f", block(xDef, empty()))
+
+        // when
+        val actualCFG = generateCFGWithSimplifiedCalls(fDef, setOf(xDef))
+
+        println(programCfgToGraphviz(actualCFG))
+
+        println(programCfgToBuilder(actualCFG))
```
Giselus, line 202: I guess you want to delete prints?

PR #334, opened 2025-01-31
Sklejka merge

mhorod:


src/main/kotlin/cacophony/controlflow/functions/GenerateHandlers.kt
```diff
@@ -14,69 +13,45 @@ fun generateCallableHandlers(
...
+                    analyzed,
+                    handlerChain,
                     callConvention,
                     variablesMap,
                     escapeAnalysis,
                 )
-            ancestorHandlers[function] = functionAncestorHandlers
+            ancestorHandlers[lambda] = handlerChain
+        } else {
+            error("Lambda expression $lambda is neither a closure nor a static function")
```
jjaworska: we can factor this `else` out if we use `require`

src/main/kotlin/cacophony/semantic/names/NameResolver.kt
```diff
@@ -240,7 +246,7 @@ private class SymbolsTable {
         }
     }
 
-    fun define(id: String, definition: Definition) {
+    fun define(id: String, definition: Definition, arity: Int? = null) {
```
jjaworska: so we introduce "arity" but it's not used yet... Why is that?

test.cac
jjaworska, line 1: we probably don't want this in main?

examples/test.cac
jjaworska, line 1: we probably don't want this on main?

jjaworska (changes requested):
After a veeery brief check, these are some minor things to look at

src/main/kotlin/cacophony/semantic/names/NameResolver.kt
mhorod: I have no idea, but good catch, removed

test.cac
mhorod: i think so

examples/test.cac
mhorod: probably yes

jjaworska (approved):
If you're not scared, let's go ahead and merge it :laughing:

